PR #1, opened 2019-03-12
Initial specs

cpanato:
- Adding initial spec manifest
- add a skeleton to we might use to create the logic to deploy all the components


cpanato:
@lieut-data @jwilander please take a look

lieut-data:
@cpanato, I'm not likely to get to this until Tuesday at the earliest -- is that ok?

jwilander:
I'm going to look at this today

pkg/controller/mattermostcluster/mattermostcluster_controller.go
```diff
@@ -100,54 +110,524 @@ func (r *ReconcileMattermostCluster) Reconcile(request reconcile.Request) (recon
...
-	reqLogger.Info("Skip reconcile: Pod already exists", "Pod.Namespace", found.Namespace, "Pod.Name", found.Name)
+	// Already exists - don't requeue
 	return reconcile.Result{}, nil
 }
 
-// newPodForCR returns a busybox pod with the same name/namespace as the cr
-func newPodForCR(cr *mattermostv1alpha1.MattermostCluster) *corev1.Pod {
-	labels := map[string]string{
-		"app": cr.Name,
+func (r *ReconcileMattermostCluster) setDefaults(mattermost *mattermostv1alpha1.MattermostCluster, reqLogger logr.Logger) error {
```
jwilander: This function doesn't actually use the pointer receiver `r *ReconcileMattermostCluster` anywhere so we can just make it a function in the package. I'd also suggest generalizing the function so that it doesn't have to be tied to a reconcile request. That way if someone wants to use this package for use outside the operator they can do so. In fact, it might even make sense to move this into the `v1alpha1` package as a method directly on the `*MattermostCluster` type. So you would just do something like this above in `Reconcile`:

```go
mattermost := &mattermostv1alpha1.MattermostCluster{}
// ... other stuff ...
reqLogger.Info("Setting defaults")
err := mattermost.SetDefaults()
if err != nil {
}
```

If we really wanted to keep the `reqLogger` around to log when a default was set, we could then have a function in this package that compares the changes made to the `MattermostCluster` then outputs what defaults were set but at this point I'm not sure if that's needed.

pkg/controller/mattermostcluster/mattermostcluster_controller.go
```diff
@@ -100,54 +110,524 @@ func (r *ReconcileMattermostCluster) Reconcile(request reconcile.Request) (recon
...
+
+	if len(mattermost.Spec.DatabaseType.Type) == 0 {
+		reqLogger.Info("Setting default Mattermost database type: " + constants.DefaultMattermostDatabaseType)
+		mattermost.Spec.DatabaseType.Type = constants.DefaultMattermostDatabaseType
+	}
+
+	return nil
+}
+
+func (r *ReconcileMattermostCluster) checkDBServiceAccount(mattermost *mattermostv1alpha1.MattermostCluster, reqLogger logr.Logger) error {
```
jwilander: To me this method does three things:
1. Creates/gets the service account spec
2. Creates the service account, if necessary
3. Updates the service account, if necessary (not implemented yet)

My guess is that we (or someone else) may want to do these at different times so I'd suggest creating three smaller functions that `checkDBServiceAccount` can call.

1. `func (c *MattermostCluster) GetServiceAccountSpec() *corev1.ServiceAccount` - this one can move into the `v1alpha1` package as it fits there better and can be used outside reconciling
2. `func (r *ReconcileMattermostCluster) createServiceAccount(mattermost *mattermostv1alpha1.MattermostCluster, reqLogger logr.Logger) error`
3. `func (r *ReconcileMattermostCluster) updateServiceAccount(mattermost *mattermostv1alpha1.MattermostCluster, reqLogger logr.Logger) error`

Then `checkDBServiceAccount` can call `createServiceAccount`/`updateServiceAccount` as needed. Both of those could also call `GetServiceAccountSpec` themselves to get the service account spec or we could have that passed in, 0/5 on that

pkg/controller/mattermostcluster/mattermostcluster_controller.go
```diff
@@ -100,54 +110,524 @@ func (r *ReconcileMattermostCluster) Reconcile(request reconcile.Request) (recon
...
+	// 	if err != nil {
+	// 		return err
+	// 	}
+	// 	_ = controllerutil.SetControllerReference(mattermost, foundServiceAccount, r.scheme)
+	// }
+
+	return nil
+}
+
+func (r *ReconcileMattermostCluster) checkDBRoleBinding(mattermost *mattermostv1alpha1.MattermostCluster, reqLogger logr.Logger) error {
```
jwilander: Similar suggestion of splitting into smaller functions as above with `checkDBServiceAccount`. We should do this for all similar methods

pkg/controller/mattermostcluster/mattermostcluster_controller.go
```diff
@@ -100,54 +110,524 @@ func (r *ReconcileMattermostCluster) Reconcile(request reconcile.Request) (recon
...
+	dbSecret := &corev1.Secret{}
+	err := r.client.Get(context.TODO(), types.NamespacedName{Name: dbSecretName, Namespace: mattermost.Namespace}, dbSecret)
+	if err != nil {
+		return "", err
+	}
+	return string(dbSecret.Data["password"]), nil
+}
+
+func (r *ReconcileMattermostCluster) checkMattermostDeployment(mattermost *mattermostv1alpha1.MattermostCluster, reqLogger logr.Logger) error {
+	//TODO: Create the logic to deploy MM including service and ingress
```
jwilander: Is this todo still needed?

jwilander:
After @lieut-data takes a look, let's hop on a call to discuss before you make any big changes @cpanato

pkg/apis/mattermost/v1alpha1/mattermostcluster_types.go
```diff
@@ -1,38 +1,77 @@
...
+
+type DatabaseType struct {
+	Type string `json:"type,omitempty"`
 }
 
 // MattermostClusterStatus defines the observed state of MattermostCluster
 // +k8s:openapi-gen=true
 type MattermostClusterStatus struct {
-	// INSERT ADDITIONAL STATUS FIELD - define observed state of cluster
 	// Important: Run "operator-sdk generate k8s" to regenerate code after modifying this file
```
lieut-data: Should we remove this comment, like above?

pkg/apis/mattermost/v1alpha1/mattermostcluster_types.go
```diff
@@ -1,38 +1,77 @@
...
 	// Add custom validation using kubebuilder tags: https://book.kubebuilder.io/beyond_basics/generating_crd.html
+
+	// Represents whether any actions on the underlaying managed objects are
```
lieut-data: s/underlaying/underlying/

pkg/controller/mattermostcluster/constants/constants.go
```diff
@@ -0,0 +1,13 @@
+package constants
+
+const (
+	// OperatorName is a operator name
```
lieut-data:
```suggestion
	// OperatorName is the name of the Mattermost operator
```

pkg/controller/mattermostcluster/constants/constants.go
```diff
@@ -0,0 +1,13 @@
...
+	OperatorName = "mattermost-operator"
+	// DefaultAmountOfPods is the default amount of Mattermost pods
+	DefaultAmountOfPods = 2
+	// DefaultMattermostImage is the default Mattermost docker image
+	DefaultMattermostImage        = "mattermost/mattermost-enterprise-edition:5.8.0"
+	DefaultMattermostDatabaseType = "mysql"
```
lieut-data: nit: propose being consistent in adding GoDoc to all fields, if we're going to have GoDoc at all in this type.

pkg/controller/mattermostcluster/mattermostcluster_controller.go
```diff
@@ -100,54 +110,524 @@ func (r *ReconcileMattermostCluster) Reconcile(request reconcile.Request) (recon
...
+			Ports: []corev1.ServicePort{mattermostPort},
+			Selector: map[string]string{
+				constants.ClusterLabel: mattermost.Name,
+			},
+			ClusterIP: corev1.ClusterIPNone,
+		},
+	}
+
+	foundService := &corev1.Service{}
+	errGet := r.client.Get(context.TODO(), types.NamespacedName{Name: mattermost.Name, Namespace: mattermost.Namespace}, foundService)
```
lieut-data: nit: Looks like if one already exists, we can bail out early. Perhaps invert this to check up front and do exactly that, and then you can outdent the code that creates one instead?

Similar feedback in the methods below.

pkg/controller/mattermostcluster/mattermostcluster_controller.go
```diff
@@ -100,54 +110,524 @@ func (r *ReconcileMattermostCluster) Reconcile(request reconcile.Request) (recon
...
+			OwnerReferences: []metav1.OwnerReference{
+				*metav1.NewControllerRef(mattermost, schema.GroupVersionKind{
+					Group:   mattermostv1alpha1.SchemeGroupVersion.Group,
+					Version: mattermostv1alpha1.SchemeGroupVersion.Version,
+					Kind:    "MattermostCluster",
+				}),
+			},
+		},
+		Spec: appsv1.DeploymentSpec{
+			Replicas: &mattermost.Spec.Replicas,
```
lieut-data: One thing that's not clear to me is how the reconciler responds to changes in data, vs. just creating them: say if the `Replicas` value increases -- it /seems/ like this code can only handle creating the deployment, and not modify the existing one.

Or can you elaborate how this works?

pkg/apis/mattermost/v1alpha1/mattermostcluster_types.go
```diff
@@ -1,38 +1,77 @@
...
+	AvailableReplicas int32 `json:"availableReplicas"`
+	// Total number of unavailable pods targeted by this Mattermost deployment.
+	UnavailableReplicas int32 `json:"unavailableReplicas"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
 
 // MattermostCluster is the Schema for the mattermostclusters API
 // +k8s:openapi-gen=true
 type MattermostCluster struct {
```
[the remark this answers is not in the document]
cpanato: will go for `ClusterInstallation`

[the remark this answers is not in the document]
cpanato: renamed

pkg/controller/mattermostcluster/mattermostcluster_controller.go
cpanato: done

pkg/controller/mattermostcluster/mattermostcluster_controller.go
cpanato: explained in the call

pkg/controller/mattermostcluster/mattermostcluster_controller.go
cpanato: remove the updateServiceAccount, because we dont have anything to update, at least for now

cpanato:
merging per discussion with @jwilander some followup tickets will be created

PR #3, opened 2019-04-10
MM-14856 Reorganize controller code

jwilander:
For this PR I narrowly focused on reorganizing and cleaning up the controller code. I was going to spend more time on this but I ended up in a bit of a rabbit hole so I felt this was a good place to break.

The major changes:

* Split into multiple files and removed the "clusterinstallation_" prefix since that's covered by the package name (no operators I looked at did this consistently so I prefer to go with something more Go idiomatic like we do in the MM server)
* Consolidated resource creation into a single function `createResource`
* Split up reconciliation of the different components (MM, mysql, minio, etc.) into different files and functions
* Generalized the functions for checking if certain resource types exist

Some future improvements:

* Logging - I'm not entirely happy with how the logging is done and I'd prefer for us to consolidate logging with the provisioning server (@lieut-data maybe a good one for you to look at)
* Potentially generalize the `create[a-zA-Z]+IfNotExists` functions into a single `createResourceIfNotExists` function (might be better to wait on this to make sure we're not over generalizing) 
* Unit tests per specific controller component (MM, mysql, minio, etc.)
* More general clean-up


pkg/controller/clusterinstallation/controller.go
```diff
@@ -0,0 +1,264 @@
...
+		reqLogger.Info("Error creating resource:", err.Error())
+		return err
+	}
+	if err := controllerutil.SetControllerReference(owner, resource, r.scheme); err != nil {
+		return err
+	}
+	return nil
+}
+
+func (r *ReconcileClusterInstallation) createServiceAccountIfNotExists(owner v1.Object, serviceAccount *corev1.ServiceAccount, reqLogger logr.Logger) error {
```
jwilander: As stated in the description it would be nice if these could be consolidated into a single `createResourceIfNotExists`

cpanato:
the logging is for kubernetes, we will have the fluentd forwarding the all the k8s logs to ES

pkg/controller/clusterinstallation/controller.go
```diff
@@ -0,0 +1,264 @@
...
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	err = r.checkMattermost(mattermost, reqLogger)
+	if err != nil {
+		return reconcile.Result{}, err
+	}
+
+	// Already exists - don't requeue
```
lieut-data: This comment confuses me a bit: it reads like an `else` case after a bunch of other conditions failed, but we reach it in all happy paths after potentially having done work. Can you help me understand the intent here?

pkg/controller/clusterinstallation/controller.go
```diff
@@ -0,0 +1,264 @@
...
+// Object combines the interfaces that all Kubernetes objects must implement.
+type Object interface {
+	runtime.Object
+	v1.Object
+}
+
+// createResource creates the provided resource and sets the owner
+func (r *ReconcileClusterInstallation) createResource(owner v1.Object, resource Object, reqLogger logr.Logger) error {
+	if err := r.client.Create(context.TODO(), resource); err != nil {
+		reqLogger.Info("Error creating resource:", err.Error())
```
lieut-data: `Info` vs `Error` or whatever the logging interface supports?

pkg/controller/clusterinstallation/controller.go
```diff
@@ -0,0 +1,264 @@
...
+
+// Object combines the interfaces that all Kubernetes objects must implement.
+type Object interface {
+	runtime.Object
+	v1.Object
+}
+
+// createResource creates the provided resource and sets the owner
+func (r *ReconcileClusterInstallation) createResource(owner v1.Object, resource Object, reqLogger logr.Logger) error {
+	if err := r.client.Create(context.TODO(), resource); err != nil {
```
lieut-data: nit: we're deviating from the code above here, and also not being "idiomatic" with respect to error handling here. Perhaps converge on the (slight more verbose):
```
err := r.client.Create(...)
if err != nil {
    // ...
}
```

pkg/controller/clusterinstallation/controller.go
```diff
@@ -0,0 +1,264 @@
...
+// Result.Requeue is true, otherwise upon completion it will remove the work from the queue.
+func (r *ReconcileClusterInstallation) Reconcile(request reconcile.Request) (reconcile.Result, error) {
+	reqLogger := log.WithValues("Request.Namespace", request.Namespace, "Request.Name", request.Name)
+	reqLogger.Info("Reconciling ClusterInstallation")
+
+	// Fetch the ClusterInstallation instance
+	mattermost := &mattermostv1alpha1.ClusterInstallation{}
+	err := r.client.Get(context.TODO(), request.NamespacedName, mattermost)
+	if err != nil {
+		if errors.IsNotFound(err) {
```
lieut-data: nit: I like the style you have later that avoids the nesting:
```
if err != nil && errors.IsNotFound(err) {
    // ...
} else if err != nil {
    // ...
}
```

pkg/controller/clusterinstallation/controller.go
```diff
@@ -0,0 +1,264 @@
...
+		reqLogger.Info("Creating service account", serviceAccount.Name)
+		return r.createResource(owner, serviceAccount, reqLogger)
+	} else if err != nil {
+		reqLogger.Error(err, "Failed to check if service account exists")
+		return err
+	}
+
+	// TODO compare found service account versus expected
+
+	reqLogger.Info("Service account reconcicled")
```
lieut-data: nit: maybe unify the comment above and this log into just:
```
    reqLogger.Debug("TOOD: reconcile existing service account")
```

Or maybe just remove the false log?

pkg/controller/clusterinstallation/controller.go
```diff
@@ -0,0 +1,264 @@
...
+	}
+
+	// TODO compare found role binding versus expected
+
+	reqLogger.Info("Role binding reconciled")
+	return nil
+}
+
+func (r *ReconcileClusterInstallation) createServiceIfNotExists(owner v1.Object, service *corev1.Service, reqLogger logr.Logger) error {
+
```
lieut-data: nit: spurious extra newline?

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -0,0 +1,45 @@
...
+}
+
+func (r *ReconcileClusterInstallation) checkMattermostIngress(mattermost *mattermostv1alpha1.ClusterInstallation, reqLogger logr.Logger) error {
+	return r.createIngressIfNotExists(mattermost, mattermost.GenerateIngress(), reqLogger)
+}
+
+func (r *ReconcileClusterInstallation) checkMattermostDeployment(mattermost *mattermostv1alpha1.ClusterInstallation, reqLogger logr.Logger) error {
+	dbPassword, err := r.getMySQLSecrets(mattermost, reqLogger)
+	if err != nil {
+		return fmt.Errorf("Error getting the database password. Err=%s", err.Error())
```
lieut-data: Use `errors.Wrap` here?

pkg/controller/clusterinstallation/controller.go
jwilander: Not a comment I added, @cpanato any insight?

cpanato: was a comment added when we bootstrap. we can remove that

pkg/controller/clusterinstallation/controller.go
jwilander: Oh actually, I think we should keep these around as is even if we don't update existing resources since creating a resource when it should have existed is still successfully "reconciling"

jwilander: (Fixed the spelling mistake though)

lieut-data: But if the resources doesn't exist, we create and return on line 184, so this log doesn't get emitted, right?

jwilander: Hmm right. Ok I'll remove it 👍

PR #4, opened 2019-04-17
add option to use external DB

cpanato:
Add option to user choose to use an external DB like AWS RDS or any other

Jira Ticket: https://mattermost.atlassian.net/browse/MM-14982


jwilander:
Adding @gabrieljackson so he can start reviewing operator code

deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
```diff
@@ -39,6 +39,10 @@ spec:
               type: object
             databaseType:
               properties:
+                externalDatabase:
+                  description: If the user want to use an external DB Can be inside
```
gabrieljackson: nit: `external DB Can be` should maybe be `external DB. This can be...`?

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -120,16 +120,45 @@ func (mattermost *ClusterInstallation) GenerateIngress() *v1beta1.Ingress {
...
+					Name: secretName,
+				},
+				Key: "externalDB",
+			},
+		}
+	} else {
+		datasourceMM := fmt.Sprintf("mysql://%s:%s@tcp(%s-mysql.%s:3306)/mattermost?charset=utf8mb4,utf8&readTimeout=30s&writeTimeout=30s", dbUser, dbPassword, mattermost.Name, mattermost.Namespace)
+		envVarDB.Value = datasourceMM
+
+		// Create the init container to check with the DB is up and running
```
gabrieljackson: nit: `check that the DB`

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -34,10 +34,21 @@ func (r *ReconcileClusterInstallation) checkMattermostIngress(mattermost *matter
...
-	dbPassword, err := r.getMySQLSecrets(mattermost, reqLogger)
-	if err != nil {
-		return errors.Wrap(err, "Error getting the database password.")
+	externalDB := false
+	var dbPassword string
+	var dbUser string
+	var err error
+	if mattermost.Spec.DatabaseType.ExternalDatabase != "" {
+		externalDB = true
+		dbPassword = ""
```
gabrieljackson: `dbPassword` is set to "" above. We shouldn't need to do it again here.

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -34,10 +34,21 @@ func (r *ReconcileClusterInstallation) checkMattermostIngress(mattermost *matter
...
+		dbUser = ""
```
gabrieljackson: Same thing as `dbPassword`.

pkg/controller/clusterinstallation/secrets.go
```diff
@@ -0,0 +1,12 @@
+package clusterinstallation
+
+import (
+	"github.com/go-logr/logr"
+
+	mattermostv1alpha1 "github.com/mattermost/mattermost-operator/pkg/apis/mattermost/v1alpha1"
+)
+
+func (r *ReconcileClusterInstallation) checkMattermostSecret(secretName, keyName, data string, mattermost *mattermostv1alpha1.ClusterInstallation, reqLogger logr.Logger) error {
+
```
gabrieljackson: nit: don't need an empty line here

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -34,10 +34,21 @@ func (r *ReconcileClusterInstallation) checkMattermostIngress(mattermost *matter
 }
 
 func (r *ReconcileClusterInstallation) checkMattermostDeployment(mattermost *mattermostv1alpha1.ClusterInstallation, reqLogger logr.Logger) error {
-	dbPassword, err := r.getMySQLSecrets(mattermost, reqLogger)
-	if err != nil {
-		return errors.Wrap(err, "Error getting the database password.")
+	externalDB := false
+	var dbPassword string
+	var dbUser string
```
gabrieljackson: nit: you can save a bit of space when you have like types. These two lines could become:
```suggestion
	var dbUser, dbPassword string
```

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -34,10 +34,21 @@ func (r *ReconcileClusterInstallation) checkMattermostIngress(mattermost *matter
 }
 
 func (r *ReconcileClusterInstallation) checkMattermostDeployment(mattermost *mattermostv1alpha1.ClusterInstallation, reqLogger logr.Logger) error {
-	dbPassword, err := r.getMySQLSecrets(mattermost, reqLogger)
-	if err != nil {
-		return errors.Wrap(err, "Error getting the database password.")
+	externalDB := false
+	var dbPassword string
```
gabrieljackson: It's a little unclear to me, but I just want to check that we don't want to set a username of password here if an external DB is used. I think that makes sense based on my understanding of what this code is doing, but I did want to check as we are setting the values if an external DB is not used.

gabrieljackson:
Looks pretty good. I just have a few questions and suggestions.

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -120,16 +120,45 @@ func (mattermost *ClusterInstallation) GenerateIngress() *v1beta1.Ingress {
...
+		initContainerCMD := fmt.Sprintf("until curl --max-time 5 http://%s-mysql.%s:3306; do echo waiting for mysql; sleep 5; done;", mattermost.Name, mattermost.Namespace)
```
lieut-data: What happens if mysql fails to come up? Is the deployment "stuck", or can it be recovered by trying again?

deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
cpanato: done

pkg/controller/clusterinstallation/mattermost.go
cpanato: if we use an external DB we should set the DSN in the secret like `mysql:/USER:PASS@tcp(HOST:3306)/mattermost?charset=utf8mb4,utf8&readTimeout=30s&writeTimeout=30s`

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
cpanato: if I'm not wrong the pod will be restarted because the readiness probe did not succeed

PR #5, opened 2019-04-17
add coveralls

cpanato:
Add coveralls to try out


cpanato:
@jwilander we can post as a status check and/or comment as well. I don't know if we want to block the PR if not reach. thoughts?

jwilander:
Yeah let's start as a non-blocking status check and go from there

cpanato:
i will check how to not make it a blocker

PR #6, opened 2019-04-23
remove check because kind already checks with --wait flag

cpanato:
when creating the kind cluster with --wait flag the command finish when the cluster is ready, so we dont need to check if the cluster is ready to use.

### DIFF
--- a/test/e2e.sh
+++ b/test/e2e.sh
@@ -53,7 +53,7 @@ run_kind() {
 
     kubectl get all --all-namespaces
 
-    echo 'Cluster ready!...'
+    echo 'Cluster ready!'
 }
 
 install_operator-sdk() {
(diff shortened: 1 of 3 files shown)


test/e2e.sh
```diff
@@ -51,13 +51,9 @@ run_kind() {
...
 
-    echo -n 'Waiting for cluster to be ready...'
-    until ! grep --quiet 'NotReady' <(kubectl get nodes --no-headers); do
-        printf '.'
-        sleep 5
-    done
-
     kubectl get all --all-namespaces
+
+    echo 'Cluster ready!...'
```
gabrieljackson: nit: we can probably remove the `...` on this message.

cpanato: yep :)

cpanato: done

PR #8, opened 2019-04-24
push docker images on master and tagged releases

cpanato:
Add the ability to push the images when we have new things on master and when we tag a release


jwilander:
FYI @gabrieljackson

PR #10, opened 2019-04-26
adding security scanner

cpanato:
### TESTING DO NOT MERGE


jasonblais:
@cpanato should we freeze this PR via `lifecycle/frozen` label?

cpanato:
waiting for an answer from who develop the microscanner , because the current orb implementation it cannot run the scanner if the docker image does not use root

PR #11, opened 2019-04-26
refactor external database to use an existing secret

cpanato:
After the implementation https://mattermost.atlassian.net/browse/MM-14982, I think a bit more and looked other operators.

and when you need to set some external information the user is responsible to create the configmap or the secret.
In our case make sense to have a secret instead of setting the DNS (which contains passwords/user) directly in the manifest.

### DIFF
--- a/pkg/controller/clusterinstallation/controller.go
+++ b/pkg/controller/clusterinstallation/controller.go
@@ -2,7 +2,6 @@ package clusterinstallation
 
 import (
 	"context"
-	"fmt"
 
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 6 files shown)


pkg/controller/clusterinstallation/controller.go
```diff
@@ -124,43 +123,34 @@ func (r *ReconcileClusterInstallation) Reconcile(request reconcile.Request) (rec
 		return reconcile.Result{}, err
 	}
 
-	if mattermost.Spec.DatabaseType.ExternalDatabase != "" {
```
jwilander: Don't we still need to check if we're using an external DB to avoid running the DB reconciliation?

cpanato: true, fixed

PR #16, opened 2019-05-14
[MM-15372] Enhanced state and upgrade support

gabrieljackson:
This change introduces the following:
 - Full 'reconciling' and 'stable' state support
 - Final health and validity checks for the mattermost installation
   k8s objects
 - Initial upgrade support for mattermost versions
 - Support for updating deployment replica count
 - Minor logging bug fixes

Initial support for https://mattermost.atlassian.net/browse/MM-15372

Other Notes:
This should be considered initial mattermost upgrade support. While the swapping out of images works, I sporadically ran into issues with DB connectivity that should be looked into further.

Regardless, the improved state support in this PR will start to give us confidence that our installations are actually running correctly instead of just creating the k8s objects and crossing our fingers.

### DIFF
--- a/deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
+++ b/deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
@@ -80,6 +80,9 @@ spec:
             Not included when requesting from the apiserver, only from the Mattermost
             Operator API itself. More info: https://github.com/kubernetes/community/blob/master/contributors/devel/api-conventions.md#spec-and-status'
           properties:
+            image:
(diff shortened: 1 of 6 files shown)


gabrieljackson:
I pushed up another branch with this commit and it seems to have failed the test here even though nothing has changed.

I don't seem to have permissions to rerun the test so I will have to push up another commit to force the tests to run again.

Thoughts @cpanato?

cpanato:
@gabrieljackson you should have permission to rerun, you can either log in into circleci or go to the checks tab in the PR and retrigger

cpanato:
@gabrieljackson retriggered

PR #19, opened 2019-05-17
add scale up/down tests

cpanato:
Adding scale up/down e2e tests

### DIFF
--- a/pkg/controller/clusterinstallation/controller_test.go
+++ b/pkg/controller/clusterinstallation/controller_test.go
@@ -48,6 +48,7 @@ func TestReconcile(t *testing.T) {
 		},
 	}
 
+<<<<<<< HEAD
 	// Register operator types with the runtime scheme.
 	apis.AddToScheme(scheme.Scheme)
 	s := scheme.Scheme
@@ -63,6 +64,19 @@ func TestReconcile(t *testing.T) {
 
 	// Create the resources that would normally be created by other operators
 	// running on the kubernetes cluster.
+=======
+	// Setup the Manager and Controller.  Wrap the Controller Reconcile function so it writes each request to a
+	// channel when it is finished.
+	mgr, err := manager.New(cfg, manager.Options{})
+	g.Expect(err).NotTo(gomega.HaveOccurred())
+	c = mgr.GetClient()
+
+	recFn, requests := SetupTestReconcile(newReconciler(mgr))
+	g.Expect(add(mgr, recFn)).NotTo(gomega.HaveOccurred())
+	defer close(StartTestManager(mgr, g))
+
+	// Create the mysql secret
+>>>>>>> 5bc1d64... fix the order
 	dbSecret := &corev1.Secret{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      ci.Name + "-mysql-root-password",
--- a/pkg/controller/clusterinstallation/controller_test.go
+++ b/pkg/controller/clusterinstallation/controller_test.go
@@ -48,7 +48,6 @@ func TestReconcile(t *testing.T) {
 		},
 	}
 
-<<<<<<< HEAD
 	// Register operator types with the runtime scheme.
 	apis.AddToScheme(scheme.Scheme)
 	s := scheme.Scheme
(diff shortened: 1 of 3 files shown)


cpanato:
will wait for https://github.com/mattermost/mattermost-operator/pull/20

test/e2e/mattermost_test.go
```diff
@@ -82,15 +82,24 @@ func mattermostScaleTest(t *testing.T, f *framework.Framework, ctx *framework.Te
...
-	// 	return err
-	// }
+	exampleMattermost.Spec.Replicas = 2
+	err = f.Client.Update(goctx.TODO(), exampleMattermost)
+	assert.Nil(t, err)
+
+	// wait for test-mm to reach 2 replicas
+	err = e2eutil.WaitForDeployment(t, f.KubeClient, namespace, "test-mm", 2, retryInterval, timeout)
+	assert.Nil(t, err)
+
```
lieut-data: Can we do our own testing here to prove that the Mattermost instances are meaningful reachable?

pkg/controller/clusterinstallation/controller_test.go
```diff
@@ -149,6 +110,45 @@ func TestReconcile(t *testing.T) {
 	}
 	defer c.Delete(context.TODO(), minioService)
 
+	// Create the ClusterInstallation object and expect the Reconcile and Deployment to be created
+	err = c.Create(context.TODO(), instance)
+	// The instance object may not be a valid object because it might be missing some required fields.
```
lieut-data: Not sure how to parse this comment -- I guess it's instructions for when the e2e test fails in the future? Would it make sense to embed this into the failure message? Also, why doesn't this fail the test?

test/e2e/mattermost_test.go
cpanato: yes, will be the next step :)

pkg/controller/clusterinstallation/controller_test.go
cpanato: me either :( not fully understand that as well

gabrieljackson: After looking into the flapping tests I encountered on another branch. I think I figured out what they were trying to say here. This is my in-flight change:

```
	// Create the ClusterInstallation object.
	err = c.Create(context.TODO(), instance)
	// The instance object may not be a valid object because it might be missing
	// some required fields. We log the error if it IsInvalid() to help with
	// troubleshooting before ensuring the error is nil.
	if apierrors.IsInvalid(err) {
		t.Logf("failed to create object, object is invalid: %v", err)
	}
	g.Expect(err).NotTo(gomega.HaveOccurred())
```

cpanato:
@lieut-data @gabrieljackson PTAL

test/e2e/mattermost_test.go
```diff
@@ -122,6 +122,17 @@ func mattermostScaleTest(t *testing.T, f *framework.Framework, ctx *framework.Te
...
+	require.NoError(t, err)
+
+	// scale down again
+	err = f.Client.Get(context.TODO(), types.NamespacedName{Name: "test-mm", Namespace: namespace}, exampleMattermost)
+	require.NoError(t, err)
+
+	exampleMattermost.Spec.Replicas = 1
+	err = f.Client.Update(context.TODO(), exampleMattermost)
+	require.NoError(t, err)
+
```
gabrieljackson: We should wait for the deployment to reach 1 again here, right?

cpanato: true, bad merge thing :(

test/e2e/mattermost_test.go
```diff
@@ -133,6 +133,13 @@ func mattermostScaleTest(t *testing.T, f *framework.Framework, ctx *framework.Te
 	err = f.Client.Update(context.TODO(), exampleMattermost)
 	require.NoError(t, err)
 
+	// wait for test-mm to reach 2 replicas
```
gabrieljackson: nit: wait for test-mm to reach 1 replica

cpanato: :(

cpanato: fixed

PR #26, opened 2019-05-31
add service type loadbalancer

cpanato:
We can use Ingress (then need to install an ingress controller) or now we can use a service type Loadbalancer, it will create a loadbalancer and assign the MM pods to receive the traffic.

Also added this information in the get the installation

for example, if you have a ingress it will return the DNS name like
```bash
$ kubectl get clusterinstallation test
NAME   STATE    IMAGE                                      VERSION   ENDPOINT
test   stable   mattermost/mattermost-enterprise-edition   5.11.0    demo.cpanato.spinmint.com
```

if using the loadbalancer will return the hostname for the ELB in AWS or the IP in the case of Azure
```bash
$ kubectl get clusterinstallation test
NAME   STATE    IMAGE                                      VERSION   ENDPOINT
test   stable   mattermost/mattermost-enterprise-edition   5.11.0    52.232.125.219
```

### DIFF
--- a/pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
+++ b/pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
@@ -60,6 +60,46 @@ func (mattermost *ClusterInstallation) SetDefaults() error {
 
 // GenerateService returns the service for Mattermost
 func (mattermost *ClusterInstallation) GenerateService() *corev1.Service {
+	svcAnnotations := map[string]string{
+		"service.alpha.kubernetes.io/tolerate-unready-endpoints": "true",
+	}
+	if mattermost.Spec.UseServiceLoadBalancer {
+		for k, v := range mattermost.Spec.ServiceAnnotations {
+			svcAnnotations[k] = v
+		}
+		return &corev1.Service{
+			ObjectMeta: metav1.ObjectMeta{
+				Labels:    ClusterInstallationLabels(mattermost.Name),
+				Name:      mattermost.Name,
+				Namespace: mattermost.Namespace,
+				OwnerReferences: []metav1.OwnerReference{
+					*metav1.NewControllerRef(mattermost, schema.GroupVersionKind{
+						Group:   SchemeGroupVersion.Group,
+						Version: SchemeGroupVersion.Version,
+						Kind:    "ClusterInstallation",
+					}),
+				},
+				Annotations: svcAnnotations,
+			},
+			Spec: corev1.ServiceSpec{
+				Ports: []corev1.ServicePort{
+					{
+						Name:       "https",
+						Port:       433,
+						TargetPort: intstr.FromString("app"),
+					},
+					{
+						Name:       "http",
+						Port:       80,
+						TargetPort: intstr.FromString("app"),
+					},
+				},
+				Selector: ClusterInstallationLabels(mattermost.Name),
+				Type:     corev1.ServiceTypeLoadBalancer,
+			},
+		}
+	}
+
 	return &corev1.Service{
 		ObjectMeta: metav1.ObjectMeta{
 			Labels:    ClusterInstallationLabels(mattermost.Name),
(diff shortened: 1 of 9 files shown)


pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -59,6 +59,21 @@ func (r *ReconcileClusterInstallation) checkMattermostService(mattermost *matter
 		update = true
 	}
 
+	// If we are using the loadBalancer the ClusterIp is imutable
```
jwilander:
```suggestion
	// If we are using the loadBalancer the ClusterIp is immutable
```

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -277,7 +320,7 @@ func (mattermost *ClusterInstallation) GenerateDeployment(dbUser, dbPassword str
 							Ports: []corev1.ContainerPort{
 								{
 									ContainerPort: 8065,
-									Name:          mattermost.Name,
+									Name:          "app",
```
gabrieljackson: Is there a reason we changed this value away from being based on the cluster installation name?

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -59,6 +59,21 @@ func (r *ReconcileClusterInstallation) checkMattermostService(mattermost *matter
 		update = true
 	}
 
+	// If we are using the loadBalancer the ClusterIp is immutable
```
gabrieljackson: I don't think we need this section at all. The following code block does a `reflect.DeepEqual` on the `Spec` and then replaces the whole thing if anything is different.

Did you run into that not doing the trick in your manual testing?

pkg/controller/clusterinstallation/utils.go
```diff
@@ -23,7 +23,7 @@ import (
 // check at the very end to ensure that everything in the installation is as it
 // should be. Over time, more types of checks should be added here as needed.
 func (r *ReconcileClusterInstallation) checkClusterInstallation(mattermost *mattermostv1alpha1.ClusterInstallation) (*mattermostv1alpha1.ClusterInstallationStatus, error) {
-	pods := &v1.PodList{
```
gabrieljackson: 👍

pkg/controller/clusterinstallation/utils.go
```diff
@@ -53,11 +53,52 @@ func (r *ReconcileClusterInstallation) checkClusterInstallation(mattermost *matt
...
+		}
+		err := r.client.List(context.TODO(), opts, svc)
+		if err != nil {
+			return nil, err
+		}
+		if len(svc.Items) > 1 {
+			return nil, fmt.Errorf("should return just one service but returned %d", len(svc.Items))
+		}
+		if svc.Items[0].Status.LoadBalancer.Ingress == nil {
+			return nil, fmt.Errorf("waiting for the Load Balancers be active")
```
gabrieljackson:
```suggestion
			return nil, fmt.Errorf("waiting for the Load Balancers to be active")
```

pkg/controller/clusterinstallation/utils.go
```diff
@@ -53,11 +53,52 @@ func (r *ReconcileClusterInstallation) checkClusterInstallation(mattermost *matt
...
+		}
+		if svc.Items[0].Status.LoadBalancer.Ingress[0].Hostname != "" {
```
gabrieljackson: Do we care if the Hostname and the IP values are blank? Should we fall back to something else in that case?

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
cpanato: this name is just a name and it is limited to 15 chars, or we truncate the name or we hardcode like this

pkg/controller/clusterinstallation/mattermost.go
cpanato: if use that without this change will be in an infinite loop updating the service

pkg/controller/clusterinstallation/utils.go
cpanato: maybe `not available` ? @gabrieljackson

gabrieljackson: That works for me. You can set `mmEndpoint := "not available"` when you initialize the variable above and then if it doesn't get overridden that will be the default.

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
gabrieljackson: Okay, this is fine for now.

pkg/controller/clusterinstallation/mattermost.go
gabrieljackson: Hmm, that's interesting. We can go with this and review it later if needed.

pkg/controller/clusterinstallation/utils.go
cpanato: done @gabrieljackson

PR #27, opened 2019-06-03
[MM-15598] add persistent volume to mysql

cpanato:
this PR adds a persistent volume to mysql

### DIFF
--- a/deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
+++ b/deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
@@ -54,6 +54,7 @@ spec:
               properties:
                 dbStorageSize:
                   description: Defines the storage size for the database. ie 50Gi
+                  pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
                   type: string
                 externalDatabaseSecret:
                   description: If the user want to use an external DB. This can be
@@ -81,6 +82,7 @@ spec:
             minioStorageSize:
               description: MinioStorageSize defines the storage size for minio. ie
                 50Gi
+              pattern: ^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$
               type: string
             nodeSelector:
               additionalProperties:
--- a/deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
+++ b/deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
@@ -53,8 +53,7 @@ spec:
             databaseType:
               properties:
                 dbStorageSize:
-                  description: DbStorageSize defines the storage size for the database.
-                    ie 50Gi
+                  description: Defines the storage size for the database. ie 50Gi
                   type: string
                 externalDatabaseSecret:
                   description: If the user want to use an external DB. This can be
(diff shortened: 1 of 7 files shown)


deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
```diff
@@ -52,6 +52,10 @@ spec:
               type: object
             databaseType:
               properties:
+                dbStorageSize:
+                  description: DbStorageSize defines the storage size for the database.
```
jwilander:
```suggestion
                  description: Defines the storage size for the database.
```

pkg/components/mysql/mysql.go
```diff
@@ -36,6 +39,21 @@ func Cluster(mattermost *mattermostv1alpha1.ClusterInstallation) *mysqlOperator.
...
+				ObjectMeta: metav1.ObjectMeta{
+					Name: mysqlName,
+				},
+				Spec: corev1.PersistentVolumeClaimSpec{
+					AccessModes: []corev1.PersistentVolumeAccessMode{
+						"ReadWriteOnce",
+					},
+					Resources: corev1.ResourceRequirements{
+						Requests: corev1.ResourceList{
+							corev1.ResourceStorage: resource.MustParse(mattermost.Spec.DatabaseType.DbStorageSize),
```
jwilander: What happens if I enter a bad value for `DbStorageSize`? Say "abc"?

cpanato: We need to add some validation in the yaml and in the code as well. will check

cpanato: the code has this check: `resource.MustParse(mattermost.Spec.DatabaseType.DbStorageSize),` if that fail the resource will not be created. maybe we can check that in the beginning and fail @jwilander

cpanato: we can add validation in the future, in 1.14 support some validation, i need to investigate how this works.
can be in a follow up?

cpanato: add a validation, if those fields are wrong it will complain

```
The ClusterInstallation "test" is invalid: []: Invalid value: map[string]interface {}{"metadata":map[string]interface {}{"uid":"104050fe-860e-11e9-8391-2638c91ce35c", "name":"test", "namespace":"default", "creationTimestamp":"2019-06-03T14:44:17Z", "annotations":map[strin
g]interface {}{"kubectl.kubernetes.io/last-applied-configuration":"{\"apiVersion\":\"mattermost.com/v1alpha1\",\"kind\":\"ClusterInstallation\",\"metadata\":{\"annotations\":{},\"name\":\"test\",\"namespace\":\"default\"},\"spec\":{\"databaseType\":{\"dbStorageSize\":\"11
GRT\"},\"ingressName\":\"demo.cpanato.spinmint.com\",\"minioStorageSize\":\"1Gi\",\"replicas\":1}}\n"}, "generation":1}, "spec":map[string]interface {}{"replicas":1, "databaseType":map[string]interface {}{"dbStorageSize":"11GRT"}, "ingressName":"demo.cpanato.spinmint.com"
, "minioStorageSize":"1Gi"}, "apiVersion":"mattermost.com/v1alpha1", "kind":"ClusterInstallation"}: validation failure list:
spec.databaseType.dbStorageSize in body should match '^([+-]?[0-9.]+)([eEinumkKMGTP]*[-+]?[0-9]*)$'
```

cpanato: done @jwilander

cpanato:
@gabrieljackson we can add an option to set the storage class. now it will use the default one

PR #31, opened 2019-06-09
adding ingress annotations

cpanato:
- similar to what we did for the service, now for the ingress

### DIFF
--- a/deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
+++ b/deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
@@ -75,6 +75,10 @@ spec:
             image:
               description: Image defines the ClusterInstallation Docker image.
               type: string
+            ingressAnnotations:
+              additionalProperties:
+                type: string
+              type: object
             ingressName:
               description: IngressName defines the name to be used when creating the
                 ingress rules
--- a/deploy/operator.yaml
+++ b/deploy/operator.yaml
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: mattermost-operator
       containers:
         - name: mattermost-operator
-          image: mattermost/mattermost-operator:latest
+          image: mattermost/mattermost-operator:test
           command:
           - mattermost-operator
           imagePullPolicy: IfNotPresent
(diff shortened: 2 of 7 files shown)


pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -135,6 +135,11 @@ func (mattermost *ClusterInstallation) GenerateService() *corev1.Service {
 
 // GenerateIngress returns the ingress for Mattermost
 func (mattermost *ClusterInstallation) GenerateIngress() *v1beta1.Ingress {
+	ingressAnnotations := map[string]string{}
```
gabrieljackson: mattermost.Spec.IngressAnnotations is already a map[string]string, so we should just set `Annotations:  mattermost.Spec.IngressAnnotations` below.

cpanato: done

deploy/operator.yaml
```diff
@@ -15,7 +15,7 @@ spec:
       serviceAccountName: mattermost-operator
       containers:
         - name: mattermost-operator
-          image: mattermost/mattermost-operator:latest
+          image: mattermost/mattermost-operator:test
```
jwilander: Can we leave this on latest?

deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
```diff
@@ -79,6 +79,10 @@ spec:
             image:
               description: Image defines the ClusterInstallation Docker image.
               type: string
+            ingressAnnotations:
+              additionalProperties:
+                type: string
+              type: object
```
jwilander: Can you run `make yaml` to update the install files?

deploy/operator.yaml
cpanato: no because this is being used for the e2e tests. can we change that to use latest instead of test. or sed the file to replace `latest` to `test` during the e2e @jwilander

deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
cpanato: done

deploy/operator.yaml
gabrieljackson: Yeah, this should be `:test`. Some more info about it here:

https://github.com/mattermost/mattermost-operator/pull/32#issuecomment-500937414

Reasoning for it here:

https://github.com/operator-framework/operator-sdk/blob/master/doc/test-framework/writing-e2e-tests.md#running-the-tests

jwilander: Ok cool, I'll just add a sed to the makefile command

PR #32, opened 2019-06-10
[MM-15977] Enhance mattermost version change behavior

gabrieljackson:
This change includes improvements to the ability of the operator
to move between mattermost versions. The operator will now run
a job of the requested image version in order to perform any
database migrations that may be required. If those complete
successfully then the deployment will be updated to use the new
mattermost image as well.

This change also improves logging and corrects an issue with `e2e`
tests not running the correct operator image.

Fixes https://mattermost.atlassian.net/browse/MM-15977

### DIFF
--- a/deploy/role.yaml
+++ b/deploy/role.yaml
@@ -107,3 +107,14 @@ rules:
   - "*"
   verbs:
   - "*"
+- apiGroups:
+  - "batch"
+  resources:
+  - jobs
+  verbs:
+  - get
+  - create
+  - list
+  - delete
+  - watch
+  - update
--- a/pkg/controller/clusterinstallation/mattermost.go
+++ b/pkg/controller/clusterinstallation/mattermost.go
@@ -4,12 +4,16 @@ import (
 	"context"
 	"fmt"
 	"reflect"
+	"time"
 
 	"github.com/go-logr/logr"
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
 	corev1 "k8s.io/api/core/v1"
 	v1beta1 "k8s.io/api/extensions/v1beta1"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
 	"k8s.io/apimachinery/pkg/util/intstr"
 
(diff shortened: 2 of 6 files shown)


deploy/role.yaml
```diff
@@ -107,3 +107,14 @@ rules:
...
+  - "batch"
+  resources:
+  - jobs
+  verbs:
+  - get
+  - create
+  - list
+  - delete
+  - watch
+  - update
```
jwilander: Can you run `make yaml` to update the install file?

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -207,6 +206,87 @@ func (r *ReconcileClusterInstallation) updateMattermostDeployment(mi *mattermost
...
+		}
+		defer func() {
+			err = r.client.Delete(context.TODO(), job)
+			if err != nil {
+				reqLogger.Error(err, "Unable to cleanup image update check job")
+			}
+		}()
+
+		// Wait up to 60 seconds for the update check job to return successfully.
+		timer := time.NewTimer(60 * time.Second)
```
jwilander: What if the upgrade job takes longer than 60s to run? We've had schema changes in the past that can take minutes (and sometimes hours) to apply when modifying large tables. We avoid making changes to those large tables but I wouldn't be surprised if we had upgrades that took 10-15min to complete in the future

deploy/role.yaml
gabrieljackson: Sure thing!

pkg/controller/clusterinstallation/mattermost.go
gabrieljackson: Interesting. I had forgotten that that was a possibility.

I think that if we want to wait any longer than 60 seconds then we shouldn't use a timer-based approach. Instead, we should start the job and leave it running while the reconcile loop checks on it over time. I can definitely develop this, but will need to do additional testing.

Do we want to merge this version and then do a follow-up PR, or wait until the next iteration is ready?

gabrieljackson:
I just noticed that running the `make yaml` process moved us back to the `:test` tag in the manifest. We need the `:test` tag in https://github.com/mattermost/mattermost-operator/blob/master/deploy/operator.yaml so we will have to do a follow-up PR to allow the `make yaml` process to deal with this correctly.

pkg/controller/clusterinstallation/mattermost.go
jwilander: I think it's fine to merge this for now since it's already better than what we have. I've created a ticket here to follow-up https://mattermost.atlassian.net/browse/MM-16276 that we can talk about next sprint planning

PR #34, opened 2019-06-12
Reorder Mattermost Service Ports

gabrieljackson:
Trying this to see if it resolves our HTTPS ELB issue.


gabrieljackson:
This is a very small change so I will merge with the single review.

PR #40, opened 2019-06-13
Set the SiteURL with an env var

jwilander:
Figured it makes sense to do this since the spec requires the ingressName to be set.


gabrieljackson:
Why have one SiteURL env var, when you could have two!?

https://github.com/mattermost/mattermost-operator/pull/38

jwilander:
Lol @cpanato beat me to it

PR #42, opened 2019-06-17
add mattermost license

cpanato:
- Add a mattermost license using the secret.

### DIFF
--- a/pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
+++ b/pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
@@ -311,8 +311,7 @@ func (mattermost *ClusterInstallation) GenerateDeployment(dbUser, dbPassword str
 	// Mattermost License
 	volumeLicense := []corev1.Volume{}
 	volumeMountLicense := []corev1.VolumeMount{}
-	if mattermost.Spec.MattermostLicenseSecret != "" {
-
+	if isLicensed {
 		envVarGeneral = append(envVarGeneral, corev1.EnvVar{
 			Name:  "MM_SERVICESETTINGS_LICENSEFILELOCATION",
 			Value: "/mattermost-license/license",
--- a/pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
+++ b/pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
@@ -173,7 +173,7 @@ func (mattermost *ClusterInstallation) GenerateIngress() *v1beta1.Ingress {
 }
 
 // GenerateDeployment returns the deployment spec for Mattermost
-func (mattermost *ClusterInstallation) GenerateDeployment(dbUser, dbPassword string, externalDB bool, minioService string) *appsv1.Deployment {
+func (mattermost *ClusterInstallation) GenerateDeployment(dbUser, dbPassword string, externalDB, isLicensed bool, minioService string) *appsv1.Deployment {
 	// Generate database config
 	envVarDB := corev1.EnvVar{
 		Name: "MM_CONFIG",
@@ -308,6 +308,33 @@ func (mattermost *ClusterInstallation) GenerateDeployment(dbUser, dbPassword str
 		},
 	}
 
+	// Mattermost License
+	volumeLicense := []corev1.Volume{}
+	volumeMountLicense := []corev1.VolumeMount{}
+	if mattermost.Spec.MattermostLicenseSecret != "" {
+
+		envVarGeneral = append(envVarGeneral, corev1.EnvVar{
+			Name:  "MM_SERVICESETTINGS_LICENSEFILELOCATION",
+			Value: "/mattermost-license/license",
+		})
+
+		volumeMountLicense = append(volumeMountLicense, corev1.VolumeMount{
+			MountPath: "/mattermost-license",
+			Name:      "mattermost-license",
+			ReadOnly:  true,
+		})
+
+		volumeLicense = append(volumeLicense, corev1.Volume{
+			Name: "mattermost-license",
+			VolumeSource: corev1.VolumeSource{
+				Secret: &corev1.SecretVolumeSource{
+					SecretName: mattermost.Spec.MattermostLicenseSecret,
+				},
+			},
+		})
+	}
+
+	// EnvVars Section
 	envVars := []corev1.EnvVar{envVarDB}
 	envVars = append(envVars, envVarMinio...)
 	envVars = append(envVars, envVarES...)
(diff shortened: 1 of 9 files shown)


pkg/controller/clusterinstallation/utils.go
```diff
@@ -112,14 +112,21 @@ func (r *ReconcileClusterInstallation) checkClusterInstallation(mattermost *matt
...
+func (r *ReconcileClusterInstallation) checkSecret(secretName, keyName, namespace string) (bool, error) {
 	foundSecret := &corev1.Secret{}
 	err := r.client.Get(context.TODO(), types.NamespacedName{Name: secretName, Namespace: namespace}, foundSecret)
 	if err != nil {
-		return errors.Wrap(err, "Error getting secret")
+		return false, errors.Wrap(err, "Error getting secret")
 	}
 
-	return nil
+	for key := range foundSecret.Data {
```
gabrieljackson: This should probably be:
```suggestion
	for _, key := range foundSecret.Data {
```

cpanato:
ready to go @gabrieljackson

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -308,6 +308,33 @@ func (mattermost *ClusterInstallation) GenerateDeployment(dbUser, dbPassword str
 		},
 	}
 
+	// Mattermost License
+	volumeLicense := []corev1.Volume{}
+	volumeMountLicense := []corev1.VolumeMount{}
+	if mattermost.Spec.MattermostLicenseSecret != "" {
+
```
gabrieljackson: nit: extra line

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -308,6 +308,33 @@ func (mattermost *ClusterInstallation) GenerateDeployment(dbUser, dbPassword str
...
+		})
+
+		volumeMountLicense = append(volumeMountLicense, corev1.VolumeMount{
+			MountPath: "/mattermost-license",
+			Name:      "mattermost-license",
+			ReadOnly:  true,
+		})
+
+		volumeLicense = append(volumeLicense, corev1.Volume{
+			Name: "mattermost-license",
```
gabrieljackson: Just checking: should this be called `license`?

I see our full path in the env var is `/mattermost-license/license`.

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -173,7 +173,7 @@ func (mattermost *ClusterInstallation) GenerateIngress() *v1beta1.Ingress {
 }
 
 // GenerateDeployment returns the deployment spec for Mattermost
-func (mattermost *ClusterInstallation) GenerateDeployment(dbUser, dbPassword string, externalDB bool, minioService string) *appsv1.Deployment {
+func (mattermost *ClusterInstallation) GenerateDeployment(dbUser, dbPassword string, externalDB, isLicensed bool, minioService string) *appsv1.Deployment {
```
gabrieljackson: We pass in `isLicensed` here, but don't actually use the value. Should we use that instead on line 314?

pkg/controller/clusterinstallation/controller.go
```diff
@@ -197,7 +197,8 @@ func (r *ReconcileClusterInstallation) Reconcile(request reconcile.Request) (rec
 
 func (r *ReconcileClusterInstallation) checkDatabase(mattermost *mattermostv1alpha1.ClusterInstallation, reqLogger logr.Logger) error {
 	if mattermost.Spec.DatabaseType.ExternalDatabaseSecret != "" {
-		return r.checkSecret(mattermost.Spec.DatabaseType.ExternalDatabaseSecret, mattermost.Namespace)
+		_, errSecret := r.checkSecret(mattermost.Spec.DatabaseType.ExternalDatabaseSecret, "externalDB", mattermost.Namespace)
```
gabrieljackson: nit:
```suggestion
		_, err := r.checkSecret(mattermost.Spec.DatabaseType.ExternalDatabaseSecret, "externalDB", mattermost.Namespace)
```

pkg/controller/clusterinstallation/controller.go
```diff
@@ -197,7 +197,8 @@ func (r *ReconcileClusterInstallation) Reconcile(request reconcile.Request) (rec
...
+		return errSecret
```
gabrieljackson: nit:
```suggestion
		return err
```

pkg/controller/clusterinstallation/utils.go
```diff
@@ -112,14 +112,21 @@ func (r *ReconcileClusterInstallation) checkClusterInstallation(mattermost *matt
...
+		if keyName == key {
+			return true, nil
+		}
+	}
+
+	msg := fmt.Sprintf("Missing required secret data. Want: %s", keyName)
+	return false, errors.Wrap(err, msg)
```
gabrieljackson: If you are choosing to return an error here if the value isn't found, then there isn't much point to also returning a bool as well.

Couldn't we just infer that if there is no error, then the secret key was found?

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
cpanato: my brain was melted yesterday, thanks for the catch. replaced

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
cpanato: this is just the name to match with the volume. but yes the path will be `/mattermost-license` to make it clear and the file is called `license`

pkg/controller/clusterinstallation/utils.go
cpanato: makes sense, done

pkg/controller/clusterinstallation/controller.go
```diff
@@ -197,7 +197,8 @@ func (r *ReconcileClusterInstallation) Reconcile(request reconcile.Request) (rec
 
 func (r *ReconcileClusterInstallation) checkDatabase(mattermost *mattermostv1alpha1.ClusterInstallation, reqLogger logr.Logger) error {
 	if mattermost.Spec.DatabaseType.ExternalDatabaseSecret != "" {
-		return r.checkSecret(mattermost.Spec.DatabaseType.ExternalDatabaseSecret, mattermost.Namespace)
+		err := r.checkSecret(mattermost.Spec.DatabaseType.ExternalDatabaseSecret, "externalDB", mattermost.Namespace)
+		return err
```
stylianosrigas:
```suggestion
		if err != nil {
			return errors.Wrap(err, "Error getting the external database secret.")				
		}
```

cpanato: thanks for the catch! fixed

cpanato:
@gabrieljackson PTAL

PR #45, opened 2019-06-27
update minio specs

cpanato:


cpanato:
@jwilander @stylianosrigas please review again, more change was needed

PR #47, opened 2019-07-03
[MM-16721] Add object matcher to check the mm deployment 

cpanato:
This PR adds the lib called k8s-objectmatcher which helps to compare resources and check if need an update or not.
Some fields in any resource (like services/deployments) are immutable and the Deepequal does not exclude that.

Adding this lib helps us to validate and apply the update if necessary

### DIFF
--- a/pkg/controller/clusterinstallation/create_resources.go
+++ b/pkg/controller/clusterinstallation/create_resources.go
@@ -3,6 +3,7 @@ package clusterinstallation
 import (
 	"context"
 
+	objectMatcher "github.com/banzaicloud/k8s-objectmatcher/patch"
 	"github.com/go-logr/logr"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
@@ -23,6 +24,13 @@ type Object interface {
 
 // createResource creates the provided resource and sets the owner
 func (r *ReconcileClusterInstallation) createResource(owner v1.Object, resource Object, reqLogger logr.Logger) error {
+	// adding the last applied annotation to use the object matcher later
+	// see: https://github.com/banzaicloud/k8s-objectmatcher
+	errMatcher := objectMatcher.DefaultAnnotator.SetLastAppliedAnnotation(resource)
+	if errMatcher != nil {
+		reqLogger.Error(errMatcher, "Error applying the annotation in the resource")
+		return errMatcher
+	}
 	err := r.client.Create(context.TODO(), resource)
 	if err != nil {
 		reqLogger.Error(err, "Error creating resource")
--- a/pkg/controller/clusterinstallation/create_resources.go
+++ b/pkg/controller/clusterinstallation/create_resources.go
@@ -26,12 +26,12 @@ type Object interface {
 func (r *ReconcileClusterInstallation) createResource(owner v1.Object, resource Object, reqLogger logr.Logger) error {
 	// adding the last applied annotation to use the object matcher later
 	// see: https://github.com/banzaicloud/k8s-objectmatcher
-	errMatcher := objectMatcher.DefaultAnnotator.SetLastAppliedAnnotation(resource)
-	if errMatcher != nil {
-		reqLogger.Error(errMatcher, "Error applying the annotation in the resource")
-		return errMatcher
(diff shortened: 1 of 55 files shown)


pkg/controller/clusterinstallation/create_resources.go
```diff
@@ -23,6 +24,13 @@ type Object interface {
 
 // createResource creates the provided resource and sets the owner
 func (r *ReconcileClusterInstallation) createResource(owner v1.Object, resource Object, reqLogger logr.Logger) error {
+	// adding the last applied annotation to use the object matcher later
+	// see: https://github.com/banzaicloud/k8s-objectmatcher
+	errMatcher := objectMatcher.DefaultAnnotator.SetLastAppliedAnnotation(resource)
```
gabrieljackson: nit: 
```suggestion
	err := objectMatcher.DefaultAnnotator.SetLastAppliedAnnotation(resource)
```

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -192,24 +192,20 @@ func (r *ReconcileClusterInstallation) checkMattermostDeployment(mattermost *mat
 // If an update is required then the deployment spec is set to:
 // - roll forward version
 // - keep active MattermostInstallation available by setting maxUnavailable=N-1
-func (r *ReconcileClusterInstallation) updateMattermostDeployment(mi *mattermostv1alpha1.ClusterInstallation, original, d *appsv1.Deployment, reqLogger logr.Logger) error {
+func (r *ReconcileClusterInstallation) updateMattermostDeployment(mi *mattermostv1alpha1.ClusterInstallation, modified, original *appsv1.Deployment, reqLogger logr.Logger) error {
```
gabrieljackson: nit: maybe the name `new` instead of `modified` makes sense here?

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -288,28 +284,19 @@ func (r *ReconcileClusterInstallation) updateMattermostDeployment(mi *mattermost
...
+		reqLogger.Error(err, "Error checking the difference in the deployment")
+		return err
 	}
 
-	if update {
-		mu := intstr.FromInt(int(mi.Spec.Replicas - 1))
-		if d.Spec.Strategy.RollingUpdate == nil {
-			d.Spec.Strategy.RollingUpdate = &appsv1.RollingUpdateDeployment{}
+	if !patchResult.IsEmpty() {
+		errMatcher := objectMatcher.DefaultAnnotator.SetLastAppliedAnnotation(modified)
```
gabrieljackson: nit:
```suggestion
		err = objectMatcher.DefaultAnnotator.SetLastAppliedAnnotation(modified)
```

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -288,28 +284,19 @@ func (r *ReconcileClusterInstallation) updateMattermostDeployment(mi *mattermost
...
-		update = true
-		d.Labels = updatedLabels
+	patchResult, err := objectMatcher.DefaultPatchMaker.Calculate(original, modified)
+	if err != nil {
+		reqLogger.Error(err, "Error checking the difference in the deployment")
+		return err
 	}
 
-	if update {
-		mu := intstr.FromInt(int(mi.Spec.Replicas - 1))
```
gabrieljackson: As noted in the comment for this method, this code would set the max unavailable value of the pod deployment to a specific value based on the replicas. The idea was to always have a sane max unavailable value even though one may not be set directly in the original deployment object.

I am only 1/5 on keeping this intact, but just note that if we remove this then we will change the rolling update behavior. Also, we should update the comment for this method so it no longer includes the max unavailable line.

cpanato: we can set in the deployment this should not change. i will do the logic there.

pkg/controller/clusterinstallation/mattermost.go
cpanato: done

cpanato:
@gabrieljackson if you have time, i've added the maxUnavailable in the deployment

PR #59, opened 2019-08-01
Mm 17077 Add BlueGreen support for Mattermost Operator

stylianosrigas:
This PR includes work for three tickets:
- https://mattermost.atlassian.net/browse/MM-17077
- https://mattermost.atlassian.net/browse/MM-17051
- https://mattermost.atlassian.net/browse/MM-17050

It introduces the support for BlueGreen deployments in the Mattermost Operator.
Currently supports:
- Moving from RollingUpdate mode to BlueGreen mode
- Moving from BlueGreen mode to RollingUpdate mode
- Changing traffic between Blue and Green deployments
- Rolling updates of non production deployments (either blue or green)
- Parameterised BlueGreen deployment

Future improvements: 
- Automatic deletion of old deployments when moving between modes
- k8s status labels for BlueGreen Deployment 
- Documentation updates


pkg/controller/clusterinstallation/bluegreen.go
```diff
@@ -0,0 +1,336 @@
...
+	"github.com/go-logr/logr"
+	"github.com/pkg/errors"
+	appsv1 "k8s.io/api/apps/v1"
+	batchv1 "k8s.io/api/batch/v1"
+
+	// batchv1 "k8s.io/api/batch/v1"
+	corev1 "k8s.io/api/core/v1"
+	v1beta1 "k8s.io/api/extensions/v1beta1"
+
+	// k8sErrors "k8s.io/apimachinery/pkg/api/errors"
```
cpanato: we can remove this

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -53,10 +53,39 @@ func (mattermost *ClusterInstallation) SetDefaults() error {
...
+		if bg.GreenIngressName == "" {
+			bg.GreenIngressName = fmt.Sprintf("green.%s", mattermost.Spec.IngressName)
+		}
+		if bg.BlueIngressName == "" {
+			bg.BlueIngressName = fmt.Sprintf("blue.%s", mattermost.Spec.IngressName)
+		}
+		if bg.GreenVersion == "" || bg.BlueVersion == "" {
+			return errors.New("Both Blue and Green deployment versions required, but not set")
+		}
+		if bg.ProductionDeployment == "" {
```
gabrieljackson: We should check for more than empty here. If the value is neither blue or green, we should error or do something.

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -515,3 +544,20 @@ func ClusterInstallationLabels(name string) map[string]string {
...
+
+// GetSelector returns the selector that should be used depending on whether blue-green is enabled or not
+func GetSelector(mattermost *ClusterInstallation) map[string]string {
+	if mattermost.Spec.BlueGreen.Enable {
+		if mattermost.Spec.BlueGreen.ProductionDeployment == "blue" {
+			return ClusterInstallationLabels(mattermost.Spec.BlueGreen.BlueInstallationName)
+		} else if mattermost.Spec.BlueGreen.ProductionDeployment == "green" {
+			return ClusterInstallationLabels(mattermost.Spec.BlueGreen.GreenInstallationName)
+		} else {
+			fmt.Print("Wrong ProductionDeployment type used. Select either blue or green")
```
gabrieljackson: This should log as opposed to using `fmt`.

pkg/controller/clusterinstallation/bluegreen_utils.go
```diff
@@ -0,0 +1,206 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+)
+
+// checkGreenInstallation checks the health and correctness of the k8s
+// objects that make up a GreenInstallation.
+//
+// NOTE: this is a vital health check. Every reconciliation loop should run this
+// check at the very end to ensure that everything in the installation is as it
+// should be. Over time, more types of checks should be added here as needed.
+func (r *ReconcileClusterInstallation) checkGreenInstallation(mattermost *mattermostv1alpha1.ClusterInstallation) (mattermostv1alpha1.ClusterInstallationStatus, error) {
```
gabrieljackson: So, we definitely need to check the blue and green installation, but I believe we should be able to use one function to check any deployment type. Just like you did with most of the create functions we should be able to pass in a variable, or simply look at the config to determine how the status check should perform.

I think this is important so that we keep one area for checking the status of the cluster installation.

gabrieljackson:
The changes that Carlos and I proposed have been submitted. Review the last commit to only see the changes I made to the original PR.

@jwilander @cpanato 

https://mattermost.atlassian.net/browse/MM-17784

PR #62, opened 2019-08-06
Update kind

cpanato:
- update kind to version 0.4.0
- update mattermost to use 5.14.0


cpanato:
it ran on my local machine, i will spend some time to see why it is failing on cci

low priority

.circleci/config.yml
```diff
@@ -49,12 +49,12 @@ jobs:
   e2e:
     environment:
       - CHANGE_MINIKUBE_NONE_USER=true
-      - K8S_VERSION=v1.13.2
+      - K8S_VERSION=v1.15.0
```
gabrieljackson: In the future, we may want to target our current latest-stable-kops version of k8s.

PR #65, opened 2019-08-12
[MM-17530] Add option for custom DB root/user password when using the mysql-operator for Mattermost operator

phillipahereza:
Resolves [#11816](https://github.com/mattermost/mattermost-server/issues/11816)


ankitrgadiya:
@phillipahereza Would it better if you can accept Secret directly?

phillipahereza:
@ankitrgadiya could you elaborate on this

> @phillipahereza Would it better if you can accept Secret directly?

PR #70, opened 2019-08-22
[MM-17902] BlueGreen enhancements

gabrieljackson:
This change does the following:
 - Moves the BlueGreen deployment values to separate Blue and Green
   AppDeployment structs. This allows us to cleanly create methods
   for a given deployment and to simplify configuration.
 - Adds BlueGreen unit tests and improves old unit tests.
 - Removes all nearly-identical code for creating and checking the
   BlueGreen resources. The original code for creating standard
   ClusterInstallation resources was made more flexible so that it
   can support standard or BlueGreen resources.
 - Resolves a bug regarding the deployment update-check job.

https://mattermost.atlassian.net/browse/MM-17902


pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -544,6 +547,38 @@ func (mattermost *ClusterInstallation) GetImageName() string {
...
+			return mattermost.Spec.BlueGreen.Green.Name
+		}
+	}
+
+	return mattermost.Name
+}
+
+// MockBaseDeployment returns the ClusterInstallation in the form of a bluegreen
+// deployment object.
+func (mattermost *ClusterInstallation) MockBaseDeployment() AppDeployment {
```
jwilander: This doesn't seem to be used anywhere, safe to remove?

PR #71, opened 2019-08-25
MM-17908: Added support for MySQL Backup

ankitrgadiya:
In this pull request, I added additional fields for Database backup in ClusterInstallation spec which are being passed to MySQL.

I also bumped the version of MySQL Operator from 0.3.0 to 0.3.2 which also resolves some inconsistencies between the vendor and upstream package.

Resolves [#11924](https://github.com/mattermost/mattermost-server/issues/11924)


ankitrgadiya:
@jwilander Sure, I will. Thanks

pkg/apis/mattermost/v1alpha1/clusterinstallation_types.go
```diff
@@ -143,6 +143,18 @@ type Database struct {
...
+	BackupSchedule string `json:"backupSchedule,omitempty"`
+	// Defines the object storage url for uploading backups.
+	// +optional
+	BackupURL string `json:"backupURL,omitempty"`
+	// Defines the backup retention policy.
+	// +optional
+	BackupRemoteDeletePolicy string `json:"backupRemoteDeletePolicy,omitempty"`
+	// Defines the secret to be used for uploading backup.
+	// +optional
+	BackupSecret string `json:"backupSecret,omitempty"`
```
gabrieljackson: A small request: could we rename this to match the syntax from the mysql operator `BackupSecretName`?

It's the only one that is slightly different so might as well make them all align.

ankitrgadiya: Yes makes sense. I will make the change :)

PR #72, opened 2019-08-27
[MM-16722] Use github.com/banzaicloud/k8s-objectmatcher for all resource reconciliation comparisons

jozuenoon:
Resolves: https://github.com/mattermost/mattermost-server/issues/11919

Some comments about changes:
- added custom annotation for k8s-objectmatcher `mattermost.com/last-applied`.
- change naming of reconciled resources to `current` and `desired` - this change is up to discussion. I found it much more readable this way and less likely to make mistake when passing down to `update` function. By function context it should be always obvious what `current` and `desired` are.
- `mysql-cluster` resource if used with objectmatcher have negative impacts on `mysql` operator, resulting in either cluster being stalled in read only mode or completely ignoring subsequent changes to spec. This happen in situation when `TestMattermost/mattermost_upgrade_test` starts. Since `db` resource have same name for every deployment `mysql-cluster` resource gets override (from `test-mm` to `test-mm2`) with new password secret and owner reference. I found that if `objectmatcher` is in use `mysql` annotation `mysql.presslabs.org/version` gets dropped, resulting in `mysql` operator ignoring any changes to resource. If this annotation is restored manually it makes some progress however cluster stays in read only mode.

Some followup ideas:
- would be nice to make `mysql-cluster` resource name variable, depending on name of mattermost deployment `<name>-db` and not using static `db` (this would probably prevent problems with objectmatcher on this resource),
- in e2e tests there is upgrade test right after scale test, the scale test creates database at default version, then immediately `test-mm2` overrides with downgraded version, however database schema doesn't change since it operates on same resource, then upgrade is performed, but it performs upgrades (if any) over already existing upgraded version. If testing schema upgrades is not intended here that may need a comment in tests, otherwise this should be fixed with variable db resource name likely also triggering some changes to e2e test,
- inspecting `mysql-operator` interactions with CRD and look through logs for any possible strange behaviors,
- remove `create<...>IfNotExists` functions since embedding them directly into reconcile functions would save one call to API Get and improve line of sight,
- add [getOrCreateMySQLSecrets](https://github.com/mattermost/mattermost-operator/blob/master/pkg/controller/clusterinstallation/mysql.go#L70) as fully checked resource on reconcile.


jwilander:
How's this going @jozuenoon ? Need any help?

pkg/controller/clusterinstallation/mysql.go
```diff
@@ -18,36 +18,38 @@ import (
...
-	foundCluster := &mysqlOperator.MysqlCluster{}
-	err = r.client.Get(context.TODO(), types.NamespacedName{Name: cluster.Name, Namespace: cluster.Namespace}, foundCluster)
+	current := &mysqlOperator.MysqlCluster{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: desired.Name, Namespace: desired.Namespace}, current)
 	if err != nil {
 		return err
 	}
 
+	// Updating mysql.cluster with objectMatcher breaks mysql operator.
+	// Only fields that are expected to be changed by mattermost-operator should be included here.
```
jozuenoon: Until other problems around database resource are resolved, mysql reconcile connot be changed to objectMatcher.

jwilander:
@cpanato @gabrieljackson what do you think about @jozuenoon's thoughts in the summary?

jwilander:
I think your changes make sense, especially regarding the mysql cluster. We should do our best to not interfere with the internal workings of the mysql operator and just reconcile the parts of the spec we are setting

For the follow-ups:

>would be nice to make mysql-cluster resource name variable, depending on name of mattermost deployment <name>-db and not using static db (this would probably prevent problems with objectmatcher on this resource),

I agree with not using a static name for the db cluster. The problem with using a variable name is that the mysql slaves use a really long name to connect to the master (see https://github.com/presslabs/mysql-operator/issues/170) which is why we started using the static name.

>e2e tests

I'll leave that to @cpanato's feedback

>inspecting mysql-operator interactions with CRD and look through logs for any possible strange behaviors

I'm all for that 👍 

>remove create<...>IfNotExists functions since embedding them directly into reconcile functions would save one call to API Get and improve line of sight,

Not sure what you mean on how this saves an API call? The reason we separated those out was so they could be reused if needed

>add getOrCreateMySQLSecrets as fully checked resource on reconcile

Makes sense to me

PR #74, opened 2019-09-01
Using downloaded operator-sdk for code generation

ankitrgadiya:
Makefile has `operator-sdk` target to pin it to a specific version for builds. However, for code generating it is calling just `operator-sdk` which will run the binary in `PATH` not the one downloaded.

This change makes it use the one downloaded in `build/` directory.


ankitrgadiya:
I make the `operator-sdk` target to use a script which gets the operator-sdk binary of specified version for suitable platform if not exist or update the existing one if version is changed.

ankitrgadiya:
I also made some changes in CI config since those lines are not required anymore as `get-operator-sdk.sh` does the work.

build/get-operator-sdk.sh
```diff
@@ -0,0 +1,39 @@
...
+if [ -f ./operator-sdk ] && ./operator-sdk version | grep "$VERSION"
+then
+	exit 0
+fi
+
+# Choose the version suitable for platform
+if [[ "$OSTYPE" == "darwin"* ]]
+then
+	URL="$BASE_URL/$VERSION/operator-sdk-$VERSION-$(uname -m)-apple-darwin"
+elif [[ "$OSTYPE" == "linux-gnu"* ]]
```
gabrieljackson: Let's default to the linux binary URL and overwrite it if darwin is detected. Right now, this logic could lead to an empty `$URL` variable depending on what `$OSTYPE` returns

ankitrgadiya: That is correct. I'll make the changes. Thank you :)

ankitrgadiya:
I made the requested changes

PR #76, opened 2019-09-06
MM 18092 - Adding Operator support for Canary deployments.

stylianosrigas:
Ticket -> https://mattermost.atlassian.net/browse/MM-18092

This PR is adding support for Canary deployments in the Mattermost operator.
To enable the operator ingress annotation should be passed as part of the manifest.
More information on the ingress annotations [here](https://kubernetes.github.io/ingress-nginx/user-guide/nginx-configuration/annotations/#canary)


cpanato:
will take some time to review, please bear with me.

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -62,10 +62,28 @@ func (mattermost *ClusterInstallation) SetDefaults() error {
 	mattermost.Spec.Minio.SetDefaults()
 	mattermost.Spec.Database.SetDefaults()
 	err := mattermost.Spec.BlueGreen.SetDefaults(mattermost)
```
gabrieljackson: We need to check if this error is nil or not before we overwrite it on the next line.

pkg/controller/clusterinstallation/canary.go
```diff
@@ -0,0 +1,33 @@
...
+)
+
+func (r *ReconcileClusterInstallation) checkCanary(mattermost *mattermostv1alpha1.ClusterInstallation, reqLogger logr.Logger) error {
+	if mattermost.Spec.Canary.Enable {
+		ingressAnnotations := map[string]string{
+			"kubernetes.io/ingress.class":                  "nginx",
+			"nginx.ingress.kubernetes.io/canary":           "true",
+			"nginx.ingress.kubernetes.io/canary-by-cookie": "canary",
+		}
+		reqLogger = reqLogger.WithValues("Reconcile", "mattermost")
```
gabrieljackson: nit: we normally set this logger name right at the top of the function, but really not a big deal.

deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
```diff
@@ -102,6 +102,31 @@ spec:
...
                   type: string
               type: object
+            canary:
+              properties:
+                canaryDeployment:
+                  description: CanaryDeployment defines the canary deployment.
+                  properties:
+                    image:
+                      description: Image defines the base Docker image that will be
+                        used for the deployment. Required when BlueGreen is enabled.
```
jwilander: We should probably update the comment on this field in the AppDeployment struct so that it doesn't mention BlueGreen now that it's being used for other things as well.

deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
```diff
@@ -102,6 +102,31 @@ spec:
...
+                      type: string
+                    ingressName:
+                      description: IngressName defines the ingress name that will
+                        be used by the deployment.
+                      type: string
```
jwilander: Is there a way we can not have this appear in the CRD while keeping it part of AppDeployment? If not, let's update the comment to mention that it is not used for canary builds.

deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
```diff
@@ -102,6 +102,31 @@ spec:
...
+                    name:
+                      description: Name defines the name of the deployment
+                      type: string
+                    version:
+                      description: Version defines the Docker image version that will
+                        be used for the deployment. Required when BlueGreen is enabled.
```
jwilander: Same for this one about removing mention of BlueGreen

pkg/apis/mattermost/v1alpha1/clusterinstallation_types.go
```diff
@@ -68,6 +71,16 @@ type ClusterInstallationSpec struct {
 	IngressAnnotations map[string]string `json:"ingressAnnotations,omitempty"`
 }
 
+// Canary defines the configuration of Canary deployment for a ClusterInstallation
+type Canary struct {
+	// Enable defines if Canary deployment will be applied.
```
jwilander:
```suggestion
	// Enable defines if a canary build will be deployed.
```

pkg/apis/mattermost/v1alpha1/clusterinstallation_types.go
```diff
@@ -68,6 +71,16 @@ type ClusterInstallationSpec struct {
...
+	// +optional
+	Enable bool `json:"enable,omitempty"`
+	// CanaryDeployment defines the canary deployment.
+	// +optional
+	CanaryDeployment AppDeployment `json:"canaryDeployment,omitempty"`
```
jwilander: Can we just name this `Deployment`? `CanaryDeployment` is a bit redundant

deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
stylianosrigas: @jwilander I was thinking of adding Canary as well. For example, `Required when BlueGreen or Canary is enabled`. This way the user will know that it is a required option when either of the two deployment types is used.

deploy/crds/mattermost_v1alpha1_clusterinstallation_crd.yaml
stylianosrigas: We cannot make it not to appear but I updated the comment to specify that this option is not for Canary deployments.

pkg/apis/mattermost/v1alpha1/clusterinstallation_types.go
stylianosrigas: Thanks for the suggestion. Fixed ;)

PR #77, opened 2019-09-14
MM-16276 Allow k8s reconcile loop to monitor ongoing updates

gigawhitlocks:
Hi everyone, this change should resolve [MM-16276](https://mattermost.atlassian.net/browse/MM-16276).

After this change, when `updateMattermostDeployment` detects that an update `Job` needs to be run, it does not wait for the `Job` to complete before returning control back to the reconciliation loop. Instead, it starts the `Job` and then returns an error each time it is called until the update succeeds. This keeps the cluster `Reconciling` until the update completes.

When the update `Job` is complete, `updateMattermostDeployment` will detect this and clean up the created `Job` and `Pod`, regardless as to whether it succeeded or failed.

This change in architecture allows update to overcome the limitation previously imposed by the 1 minute timeout.

### DIFF
--- a/pkg/controller/clusterinstallation/mattermost.go
+++ b/pkg/controller/clusterinstallation/mattermost.go
@@ -259,9 +259,9 @@ func (r *ReconcileClusterInstallation) updateMattermostDeployment(mi *mattermost
 		alreadyRunning, err := r.fetchRunningUpdateJob(mi, reqLogger)
 		if err != nil && k8sErrors.IsNotFound(err) {
 			reqLogger.Info("Launching update job")
-			if err := r.launchUpdateJob(mi, new, imageName, reqLogger); err != nil {
+			if er := r.launchUpdateJob(mi, new, imageName, reqLogger); er != nil {
 				reqLogger.Error(err, "Launching update job failed")
-				return err
+				return er
 			} else {
 				return errors.New("Began update job")
 			}
--- a/pkg/controller/clusterinstallation/mattermost.go
+++ b/pkg/controller/clusterinstallation/mattermost.go
@@ -4,7 +4,6 @@ import (
 	"context"
 	"fmt"
 	"reflect"
-	"time"
 
 	objectMatcher "github.com/banzaicloud/k8s-objectmatcher/patch"
 	"github.com/go-logr/logr"
@@ -15,10 +14,14 @@ import (
 	v1beta1 "k8s.io/api/extensions/v1beta1"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
 	"k8s.io/apimachinery/pkg/types"
+	k8sClient "sigs.k8s.io/controller-runtime/pkg/client"
 
 	mattermostv1alpha1 "github.com/mattermost/mattermost-operator/pkg/apis/mattermost/v1alpha1"
 )
+
+const updateName = "mattermost-update-check"
 
 func (r *ReconcileClusterInstallation) checkMattermost(mattermost *mattermostv1alpha1.ClusterInstallation, reqLogger logr.Logger) error {
 	reqLogger = reqLogger.WithValues("Reconcile", "mattermost")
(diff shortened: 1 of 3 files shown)


gigawhitlocks:
Still investigating the CI failure

gigawhitlocks:
The e2e tests appear to be failing because the scale test fails to start the second replica of the Mattermost Deployment. Still investigating why my change is causing this to occur.

jwilander:
@cpanato @gabrieljackson any insight that might help @gigawhitlocks ?

gigawhitlocks:
After the test calls `Update` I have noticed that the spec for the deployment, viewed through `kubectl`, still only shows 1 replica requested. Editing this manually makes the test proceed. Haven't yet determined why this is occurring.

gigawhitlocks:
The tests were failing due to a bug in my patch where I had inadvertently restricted manifest updates to occurring only after Mattermost version upgrades. This has been resolved and @gabrieljackson's comments should also be addressed at this time. Let me know if there's anything else that needs addressing.

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -223,75 +256,60 @@ func (r *ReconcileClusterInstallation) updateMattermostDeployment(mi *mattermost
...
-						Labels: map[string]string{"app": updateName},
-					},
-					Spec: *new.Spec.Template.Spec.DeepCopy(),
-				},
-			},
+		alreadyRunning, err := r.fetchRunningUpdateJob(mi, reqLogger)
+		if err != nil && k8sErrors.IsNotFound(err) {
+			reqLogger.Info("Launching update job")
+			if err = r.launchUpdateJob(mi, new, imageName, reqLogger); err != nil {
+				reqLogger.Error(err, "Launching update job failed")
```
gabrieljackson: Instead of logging these errors and then returning them, let's just return them with their full context. i.e. `return errors.Wrap(err, "Launching update job failed")`.

The idea is we add Info logs as needed throughout the code, but the returned errors will be logged so we can clean things up by not also separately logging them.

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -223,75 +256,60 @@ func (r *ReconcileClusterInstallation) updateMattermostDeployment(mi *mattermost
...
+			}
+			return errors.New("Began update job")
 		}
 
-		// Override values for job-specific behavior.
-		job.Spec.Template.Spec.RestartPolicy = corev1.RestartPolicyNever
-		for i := range job.Spec.Template.Spec.Containers {
-			job.Spec.Template.Spec.Containers[i].Command = []string{"mattermost", "version"}
+		if err != nil {
+			reqLogger.Error(err, "Error trying to determine if an update job already is running")
```
gabrieljackson: Same here

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -223,75 +256,60 @@ func (r *ReconcileClusterInstallation) updateMattermostDeployment(mi *mattermost
...
+			reqLogger.Info(fmt.Sprintf("Deleting %d pods", len(podList.Items)))
+			for _, p := range podList.Items {
+				reqLogger.Info(fmt.Sprintf("Deleting pod %s/%s", p.Namespace, p.Name))
+				err = r.client.Delete(context.TODO(), &p)
 				if err != nil {
-					continue
-				}
-				if foundJob.Status.Failed > 0 {
-					return errors.New("Upgrade image job check failed")
+					reqLogger.Error(err, "Problem deleting pod %s", p)
```
gabrieljackson: Looks like we probably want a `fmt.Sprintf` here?

gigawhitlocks: Oh good catch here especially

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -223,75 +256,58 @@ func (r *ReconcileClusterInstallation) updateMattermostDeployment(mi *mattermost
...
-					Spec: *new.Spec.Template.Spec.DeepCopy(),
-				},
-			},
+		alreadyRunning, err := r.fetchRunningUpdateJob(mi, reqLogger)
+		if err != nil && k8sErrors.IsNotFound(err) {
+			reqLogger.Info("Launching update job")
+			if err = r.launchUpdateJob(mi, new, imageName, reqLogger); err != nil {
+				return errors.Wrap(err, "Launching update job failed")
+			}
+			return errors.New("Began update job")
```
jwilander: This shouldn't be an error, right?

gigawhitlocks: It should, actually, so that the reconcile loop remains in the Reconciling state. If we return nil, the loop may go back to Stable, which we don't want yet. Am I overlooking a better way to handle that state signalling?

gigawhitlocks: Alterntatively I could also leave a comment there for the sake of clarity since it is a little counterintuitive.

jwilander: Nope, actually looking at it again you've got it right 👍 I do think a comment to clarify would be useful to add

gabrieljackson: Feel free to modify the comment in a future PR.

PR #75, opened 2019-09-03
Add option to restore a MySQL Mattermost Database when deploying a new Mattermost installation

cpanato:
This adds the option the user restore an existing Mattermost database when requesting a new mattermost installation

- the user need to create the back/restore secret like
```
apiVersion: v1
kind: Secret
metadata:
  name: backrestore
type: Opaque
stringData:
  AWS_ACCESS_KEY_ID: XXXXXXX
  AWS_SECRET_ACCESS_KEY: XXXX
  AWS_REGION: us-east-1
  S3_PROVIDER: AWS
```

PR #78, opened 2019-09-16
[MM-18092] Add new controller for restore MySQL DB

cpanato:
This PR is an upgrade of https://github.com/mattermost/mattermost-operator/pull/75
This adds a new controler/api to handle the restore of the MySQL DB.

Conditions to use that:
1. have a existing clusterinstallation
2. have a mysql backup using the percona xtrabckp

Flow:
User create a new clusterinstallation when this is stable they can create the restore manifest setting the required fields and the clusterinstallation name
This will trigger the following actions:
- scale the MySQL DB to zero
- delete the PVC
- set the new credentials for the db
- set back the original # of mysql replicas

We need to scale to zero in order to be able to delete the persistence volume and then we can set the mysql to use the restore data for reference can check here https://github.com/presslabs/mysql-operator/issues/390#issuecomment-529435194

during the restoring flow we can get information from the CRD and this will look like:
for restoring:
```
NAME      STATE      ORIGINAL DB REPLICAS
restore   restoring   1
```

when it is complete
```
NAME      STATE      ORIGINAL DB REPLICAS
restore   finished   1
```

This is an initial PR all feedback is welcome and maybe we create new follow-up PRs to address issues.

/cc @stylianosrigas @gabrieljackson @jwilander 

TODO:
- add unit tests
- add e2e tests


cpanato:
this was what I got in my mind @gabrieljackson it is not the best solution and far away from your implementation skills :)

pkg/controller/clusterinstallation/mysql.go
```diff
@@ -86,10 +86,13 @@ func (r *ReconcileClusterInstallation) getOrCreateMySQLSecrets(mattermost *matte
 			"DATABASE":      []byte("mattermost"),
 		}
 
-		return userPassword, r.createResource(mattermost, dbSecret, reqLogger)
+		return []string{"mmuser", userPassword, "mattermost"}, r.createResource(mattermost, dbSecret, reqLogger)
 	} else if err != nil {
 		reqLogger.Error(err, "Failed to check if mysql secret exists")
-		return "", err
+		return []string{}, err
```
jwilander: Can we always return an array of three strings, e.g. `[]string{"", "" ,""}`? Normally I'd say we should return nil for the first value any time there is an error but since line 89 could return a valid array with an error then I think it's better to be consistent and always return an array of length 3

pkg/controller/mattermostrestoredb/mattermostrestoredb_controller.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	// Check if this Mattermost ClusterInstallation exists
+	clusterInstallation := &mattermostv1alpha1.ClusterInstallation{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: restoreMM.Spec.MattermostClusterName, Namespace: restoreMM.Namespace}, clusterInstallation)
+	if err != nil && errors.IsNotFound(err) {
+		r.setFailed()
+		status := restoreMM.Status
+		status.State = r.state
+		//TODO: add reason and inform need to delete/apply when a clusterinstallation is ready.
```
jwilander: Can you create a ticket for this?

pkg/controller/mattermostrestoredb/mattermostrestoredb_controller.go
```diff
@@ -0,0 +1,234 @@
...
+	if len(pods.Items) != 0 {
+		return reconcile.Result{RequeueAfter: time.Second * 3}, fmt.Errorf("Waiting for MySQL Statefulset pods scale to 0")
+	}
+
+	reqLogger.Info("MySQL Statefulset are scaled down. Will continue the restore process", "ReadyReplicas", statefulset.Status.ReadyReplicas, "CurrentReplicas", statefulset.Status.CurrentReplicas, "OriginalReplicas", restoreMM.Status.OriginalDBReplicas)
+
+	// Removing all the PVC for MySQL to be able to apply the restore
+	for i := 0; i < int(restoreMM.Status.OriginalDBReplicas); i++ {
+		persistentVolumeClaim := &corev1.PersistentVolumeClaim{}
+		dbPersistenVolClaim := fmt.Sprintf("data-db-mysql-%d", i)
```
jwilander: s/dbPersistenVolClaim/dbPersistentVolClaim/

pkg/controller/clusterinstallation/mysql.go
cpanato: done

pkg/controller/mattermostrestoredb/mattermostrestoredb_controller.go
cpanato: done

pkg/controller/mattermostrestoredb/mattermostrestoredb_controller.go
cpanato: done:  MM-18633

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -155,11 +155,11 @@ func (r *ReconcileClusterInstallation) checkMattermostDeployment(mattermost *mat
 		}
 		externalDB = true
 	} else {
-		dbPassword, err = r.getOrCreateMySQLSecrets(mattermost, reqLogger)
+		// dbData is a []string -> { DBUserName, DBUserPassword, DBName }
```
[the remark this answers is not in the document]
cpanato: done

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -223,7 +223,7 @@ func (mattermost *ClusterInstallation) GenerateIngress(name, ingressName string)
 }
 
 // GenerateDeployment returns the deployment spec for Mattermost
-func (mattermost *ClusterInstallation) GenerateDeployment(deploymentName, ingressName, containerImage, dbUser, dbPassword string, externalDB, isLicensed bool, minioService string) *appsv1.Deployment {
+func (mattermost *ClusterInstallation) GenerateDeployment(deploymentName, ingressName, containerImage string, dbData []string, externalDB, isLicensed bool, minioService string) *appsv1.Deployment {
```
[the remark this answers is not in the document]
cpanato: done

pkg/controller/mattermostrestoredb/mattermostrestoredb_controller.go
```diff
@@ -0,0 +1,234 @@
...
+	// Check if this Mattermost ClusterInstallation exists
+	clusterInstallation := &mattermostv1alpha1.ClusterInstallation{}
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: restoreMM.Spec.MattermostClusterName, Namespace: restoreMM.Namespace}, clusterInstallation)
+	if err != nil && errors.IsNotFound(err) {
+		r.setFailed()
+		status := restoreMM.Status
+		status.State = r.state
+		//TODO: add reason and inform need to delete/apply when a clusterinstallation is ready. JIRA MM-18633
+		err = r.updateStatus(restoreMM, status, reqLogger)
+		reqLogger.Error(err, "Mattermost Installation not found. Create a ClusterInstallation first", "Namespace", restoreMM.Namespace, "ClusterInstallation.Name", restoreMM.Spec.MattermostClusterName, "RestoreDB.name", restoreMM.Name)
```
gabrieljackson: By the time we log here, we have changed the value of `err` on line 116. We should log before setting the status and then log again after if another error occurs.

pkg/controller/mattermostrestoredb/mattermostrestoredb_controller.go
```diff
@@ -0,0 +1,234 @@
...
+	// Fetch the MattermostRestoreDB instance
+	restoreMM := &mattermostv1alpha1.MattermostRestoreDB{}
+	err := r.client.Get(context.TODO(), request.NamespacedName, restoreMM)
+	if err != nil {
+		if errors.IsNotFound(err) {
+			// Request object not found, could have been deleted after reconcile request.
+			// Owned objects are automatically garbage collected. For additional cleanup logic use finalizers.
+			// Return and don't requeue
+			r.setRestoring()
+			return reconcile.Result{}, nil
```
gabrieljackson: Should we set `Requeue: false` here?

pkg/controller/mattermostrestoredb/mattermostrestoredb_controller.go
```diff
@@ -0,0 +1,234 @@
...
+	err = r.client.Get(context.TODO(), types.NamespacedName{Name: restoreMM.Spec.MattermostClusterName, Namespace: restoreMM.Namespace}, clusterInstallation)
+	if err != nil && errors.IsNotFound(err) {
+		r.setFailed()
+		status := restoreMM.Status
+		status.State = r.state
+		//TODO: add reason and inform need to delete/apply when a clusterinstallation is ready. JIRA MM-18633
+		err = r.updateStatus(restoreMM, status, reqLogger)
+		reqLogger.Error(err, "Mattermost Installation not found. Create a ClusterInstallation first", "Namespace", restoreMM.Namespace, "ClusterInstallation.Name", restoreMM.Spec.MattermostClusterName, "RestoreDB.name", restoreMM.Name)
+		return reconcile.Result{Requeue: false}, err
+	} else if err != nil {
```
gabrieljackson: We should probably log this error and any error later from trying to update the status.

pkg/controller/mattermostrestoredb/mattermostrestoredb_controller.go
```diff
@@ -0,0 +1,234 @@
...
+		return reconcile.Result{}, err
+	}
+
+	reqLogger.Info("Restore complete", "restoreMM.Namespace", restoreMM.Namespace, "restoreMM.Name", restoreMM.Name)
+	r.setFinished()
+	status := restoreMM.Status
+	status.State = r.state
+	err = r.updateStatus(restoreMM, status, reqLogger)
+	if err != nil {
+		reqLogger.Error(err, "failed to update the clusterinstallation status")
```
gabrieljackson: If we fail to set the status, it looks like we won't retry? Are we okay not returning an error here?

pkg/controller/clusterinstallation/mysql.go
```diff
@@ -67,7 +67,10 @@ func (r *ReconcileClusterInstallation) createMySQLClusterIfNotExists(mattermost
 	return nil
 }
 
-func (r *ReconcileClusterInstallation) getOrCreateMySQLSecrets(mattermost *mattermostv1alpha1.ClusterInstallation, reqLogger logr.Logger) (string, error) {
+// getOrCreateMySQLSecrets get or create the MySQL secrets used by MySQL Operator to spin the cluster and
+// also used by mattermost to get the credentials to use to access the DB.
+// dbData is a []string -> { DBUserName, DBUserPassword, DBName }
```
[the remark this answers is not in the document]
cpanato: done

pkg/controller/mattermostrestoredb/mattermostrestoredb_controller.go
gabrieljackson: Right, sorry missed that.

pkg/controller/mattermostrestoredb/mattermostrestoredb_controller.go
gabrieljackson: Yeah, just like the main clusterinstallation controller, we should return the error if the updateStatus fails. That way, we will reconcile again and will try to update the status until it works.

If we don't make sure the status is updated then we could get into a state where everything is finished, but we don't have status reflecting it.

pkg/controller/clusterinstallation/mysql.go
```diff
@@ -86,10 +96,41 @@ func (r *ReconcileClusterInstallation) getOrCreateMySQLSecrets(mattermost *matte
...
+		userPassword: string(dbSecret.Data["PASSWORD"]),
+		dbName:       string(dbSecret.Data["DATABASE"]),
+	}
+
+	return dbInfo, nil
+}
+
+func (db *databaseInfo) Valid() error {
+	if db.userName == "" {
+		return errors.New("username should be empty")
```
gabrieljackson:
```suggestion
		return errors.New("database username shouldn't be empty")
```

pkg/controller/clusterinstallation/mysql.go
```diff
@@ -86,10 +96,41 @@ func (r *ReconcileClusterInstallation) getOrCreateMySQLSecrets(mattermost *matte
...
+	}
+	if db.userPassword == "" {
+		return errors.New("password should be empty")
```
gabrieljackson:
```suggestion
		return errors.New("database password shouldn't be empty")
```

pkg/controller/clusterinstallation/mysql.go
```diff
@@ -86,10 +96,41 @@ func (r *ReconcileClusterInstallation) getOrCreateMySQLSecrets(mattermost *matte
...
 	}
-	return string(dbSecret.Data["PASSWORD"]), nil
+	if db.dbName == "" {
+		return errors.New("database name should be empty")
```
gabrieljackson:
```suggestion
		return errors.New("database name shouldn't be empty")
```

pkg/controller/mattermostrestoredb/mattermostrestoredb_controller.go
cpanato: lets do it in a followup

PR #80, opened 2019-09-20
add support for being able to use existing secret for MySQL

Sheshagiri:
This pr will address the requirement described in https://github.com/mattermost/mattermost-server/issues/11816

Fixes https://github.com/mattermost/mattermost-server/issues/11816


cpanato:
@Sheshagiri can you please update your PR with latest master? thanks!

gabrieljackson:
@Sheshagiri, would you mind merging master into your branch again and pushing it up? It looks like `e2e` tests are not passing.

pkg/apis/mattermost/v1alpha1/clusterinstallation_types.go
```diff
@@ -180,6 +180,10 @@ type Database struct {
 	BackupRemoteDeletePolicy string `json:"backupRemoteDeletePolicy,omitempty"`
 	// Defines the secret to be used for uploading/restoring backup.
 	// +optional
+	BackupSecretName string `json:"backupSecretName,omitempty"`
+	// Optionally enter the name of already existing Secret
+	// +optional
+	Secret                  string `json:"secret,omitempty"`
```
cpanato: I think this name is not a good one, maybe we can change that to something like `ExistingDBSecret`

pkg/components/mysql/mysql.go
```diff
@@ -56,7 +56,24 @@ func Cluster(mattermost *mattermostv1alpha1.ClusterInstallation) *mysqlOperator.
...
 		mysql.Spec.InitBucketSecretName = mattermost.Spec.Database.BackupRestoreSecretName
+	} else if mattermost.Spec.Database.Secret != "" {
+		mysql.Spec.SecretName = mattermost.Spec.Database.Secret
 	}
 
 	return mysql
 }
+
+// Secret returns the secret name created to use together with MySQL deployment
+func Secret(mattermost *mattermostv1alpha1.ClusterInstallation) *corev1.Secret {
```
cpanato: same for the function name, if i look this what this should do?

cpanato:
Thanks for your work here, this is going in a nice direction, just a few comments.

Sheshagiri:
> Thanks for your work here, this is going in a nice direction, just a few comments.

I'll address these over the weekend.

pkg/apis/mattermost/v1alpha1/clusterinstallation_types.go
Sheshagiri: since this is already in Database struct I renamed it to `ExistingSecret`

pkg/components/mysql/mysql.go
Sheshagiri: looks like this function is no longer required because of some refactoring on master so removed this.

Sheshagiri:
@cpanato there are no tests for workflows like this(using externalDB, external minio)
you think there should be tests for this?

pkg/components/mysql/mysql.go
```diff
@@ -56,6 +55,8 @@ func Cluster(mattermost *mattermostv1alpha1.ClusterInstallation) *mysqlOperator.
 	if mattermost.Spec.Database.InitBucketURL != "" && mattermost.Spec.Database.BackupRestoreSecretName != "" {
 		mysql.Spec.InitBucketURL = mattermost.Spec.Database.InitBucketURL
 		mysql.Spec.InitBucketSecretName = mattermost.Spec.Database.BackupRestoreSecretName
+	} else if mattermost.Spec.Database.ExistingSecret != "" {
```
gabrieljackson: I think this should be a separate `If` statement. In other words, we could use an existing secret as well as define backups.

Is that correct @cpanato?

pkg/controller/clusterinstallation/mysql.go
```diff
@@ -76,12 +76,30 @@ func (r *ReconcileClusterInstallation) createMySQLClusterIfNotExists(mattermost
...
+	// Check if the custom(existing) MySQL secret is provided
+	if mattermost.Spec.Database.ExistingSecret != "" {
+		if err := r.client.Get(context.TODO(), types.NamespacedName{
+			Namespace: mattermost.Namespace,
+			Name:      mattermost.Spec.Database.ExistingSecret,
+		}, dbSecret); err != nil {
+			return dbInfo, errors.Wrap(err, "unable to locate custom/existing MySQL secret")
+		}
+		dbInfo = databaseInfo{
+			userName:     string(dbSecret.Data["USER"]),
```
gabrieljackson: I think we should check that these three secret values are not empty before returning. If any of them are empty, we should return an error.

pkg/apis/mattermost/v1alpha1/clusterinstallation_types.go
```diff
@@ -180,6 +180,10 @@ type Database struct {
 	BackupRemoteDeletePolicy string `json:"backupRemoteDeletePolicy,omitempty"`
 	// Defines the secret to be used for uploading/restoring backup.
 	// +optional
+	BackupSecretName string `json:"backupSecretName,omitempty"`
+	// Optionally enter the name of already existing Secret
```
gabrieljackson: We may want to be more specific here on the description as we already have a few secrets that can be defined in the spec. Maybe something like:
```suggestion
	// Optionally enter the name of an already existing Secret for use by the MySQL operator
```

pkg/controller/clusterinstallation/mysql.go
Sheshagiri: I made use of the existing `dbInfo.Valid()` which will return an error if any of the secret values are empty. 
https://github.com/mattermost/mattermost-operator/blob/c85abeaa8d19960b2eb9fbd66e7107b8cd9a3551/pkg/controller/clusterinstallation/mysql.go#L94-L99

pkg/components/mysql/mysql.go
Sheshagiri: It makes sense, separated the check now :)

pkg/controller/clusterinstallation/mysql.go
gabrieljackson: Oh, missed that. Looks good!

Sheshagiri:
sure, I will try to make the suggested changes over the weekend :)

gabrieljackson:
Hey @Sheshagiri, just checking in on this. Do you still think you will make the proposed changes?

If not we can merge as is and someone else can do a follow-up PR. Just let us know!

Sheshagiri:
I will make those changes in this pr:)

Sheshagiri:
@gabrieljackson I closely looked at what you suggested and indeed that is how I implemented this. But then @cpanato [suggested that](https://github.com/mattermost/mattermost-operator/pull/80#discussion_r332424679) we rename the `Secret` to something meaningful. Based on that code review I renamed that to `ExistingSecret ` those changes are available here https://github.com/mattermost/mattermost-operator/pull/80/commits/550fe8f7490addd04708f5b36e292ead3c1fc7a5. I'm okay with bringing those changes back if @cpanato is comfortable.

cc: @gabrieljackson @cpanato

gabrieljackson:
@Sheshagiri apologies for the confusion on this. You are right that we requested the name change to `ExistingSecret`. Since then, @cpanato and I have chatted along with the rest of the team and we realized that it makes sense to have one database secret that covers both external databases as well as using custom credentials for the mysql operator database if that is chosen instead. The secret already exists here:
https://github.com/mattermost/mattermost-operator/blob/8b63d5aa0d8c830f463d0996a59de56e9850b864/pkg/apis/mattermost/v1alpha1/clusterinstallation_types.go#L155

If you could rename this secret to `Secret` and modify the operator to use it in both situations I noted above, it would be appreciated. Thanks!

cpanato:
all good we discussed internally and sounds ok to me, sorry for the delay

PR #83, opened 2019-10-03
MM-17587 Allow using external MinIO/S3 instances

jwilander:
https://mattermost.atlassian.net/browse/MM-17587

### DIFF
--- a/pkg/apis/mattermost/v1alpha1/clusterinstallation_types.go
+++ b/pkg/apis/mattermost/v1alpha1/clusterinstallation_types.go
@@ -141,7 +141,7 @@ type Minio struct {
 	ExternalBucket string `json:"externalBucket,omitempty"`
 	// Optionally enter the name of already existing secret.
 	// Secret should have two values: "accesskey" and "secretkey".
-	// Required when 'ExternalURL' is set.
+	// Required when "ExternalURL" is set.
 	// +optional
 	Secret string `json:"secret,omitempty"`
 }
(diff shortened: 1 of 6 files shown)


pkg/apis/mattermost/v1alpha1/clusterinstallation_types.go
```diff
@@ -133,7 +133,15 @@ type Minio struct {
...
-	// Optionally enter the name of already existing Secret
+	// Set to use an external MinIO deployment or S3. Must also set 'Secret' and 'ExternalBucket'.
+	// +optional
+	ExternalURL string `json:"externalURL,omitempty"`
+	// Set to the bucket name of your external MinIO or S3.
+	// +optional
+	ExternalBucket string `json:"externalBucket,omitempty"`
+	// Optionally enter the name of already existing secret.
+	// Secret should have two values: "accesskey" and "secretkey".
+	// Required when 'ExternalURL' is set.
```
gabrieljackson: nit: the formatting is strange for our generated comments with this single quote
```suggestion
	// Required when "ExternalURL" is set.
```

jwilander:
@gabrieljackson 

1. Definitely think this makes sense, good suggestion I'll make that update 👍 
2. The only problem with placing the URL and bucket in the secret is that before it was possible to set your own secrets for MinIO while still using the MinIO cluster the operator spins up. If we were to put them all in the secret together we couldn't do that be cause our only check on whether we'd be using an external S3/MinIO would be if the secret is set or not. Does that make sense or am I missing something?

gabrieljackson:
Your point about the secrets also being used for Minio operator installations is true. We could probably get around that, but it probably isn't worth the effort at this time.

PR #87, opened 2019-10-11
Bump minio

cpanato:
- minio remove the version field in the spec and replace that to image
- also they change the sufix for the service name


pkg/components/minio/minio.go
```diff
@@ -38,7 +38,6 @@ func Instance(mattermost *mattermostv1alpha1.ClusterInstallation) *minioOperator
 			},
 		},
 		Spec: minioOperator.MinIOInstanceSpec{
-			Version:     "RELEASE.2018-11-22T02-51-56Z",
```
gabrieljackson: I don't see us setting the MinIO version anywhere else now. Is that an option they removed entirely?

cpanato: the replace to the Image field, but also they have a default value.

PR #93, opened 2019-10-25
Check Mattermost pod readiness

gabrieljackson:
The final health check for the Mattermost pods didn't check for
readiness. This could lead to the cluster installation passing
the operator check and being marked as stable, only to fail future
deployment health checks and then be restarted.

https://mattermost.atlassian.net/browse/MM-19298

### DIFF
--- a/pkg/controller/clusterinstallation/utils.go
+++ b/pkg/controller/clusterinstallation/utils.go
@@ -107,6 +107,21 @@ func (r *ReconcileClusterInstallation) checkClusterInstallation(name, imageName,
 		if pod.Spec.Containers[0].Image != imageName {
 			return status, fmt.Errorf("mattermost pod %s is running incorrect image", pod.Name)
 		}
+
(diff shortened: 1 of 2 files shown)


pkg/controller/clusterinstallation/utils.go
```diff
@@ -107,6 +107,21 @@ func (r *ReconcileClusterInstallation) checkClusterInstallation(name, imageName,
 		if pod.Spec.Containers[0].Image != imageName {
 			return status, fmt.Errorf("mattermost pod %s is running incorrect image", pod.Name)
 		}
+
+		podIsReady := false
+		for _, condition := range pod.Status.Conditions {
+			if condition.Type == corev1.PodReady {
+				if condition.Status == corev1.ConditionTrue {
+					podIsReady = true
```
jwilander: nit: we could `break` out of the loop here, right?

PR #96, opened 2019-11-02
[MM-16986]Find a long term solution to MySQL cluster name length

jozuenoon:
Initial idea can go under review.

However there are 2 things I want to look into later:

1. Try to add `objectMatcher` for MySQL CRD,
2. Make some improvements around e2e tests.

Jira: https://mattermost.atlassian.net/browse/MM-16986


jwilander:
Approach looks good to me. Have you confirmed that it works with read replicas and that they are able to connect to the master?

gabrieljackson:
Yeah, I like how this looks as well. It is flexible enough that it should work in all cases.

Adding `objectMatcher` would also be a really nice improvement. Let us know how that goes.

jwilander:
@gabrieljackson ready for us to review

test/e2e/mattermost_test.go
```diff
@@ -2,9 +2,12 @@ package e2e
 
 import (
 	"context"
+	"fmt"
 	"testing"
 	"time"
 
+	"github.com/mattermost/mattermost-operator/pkg/components/utils"
+
```
gabrieljackson: nit: extra space

test/e2e/utils.go
```diff
@@ -7,13 +7,35 @@ import (
...
+		if mysql.Status.ReadyNodes == replicas {
+			return true, nil
+		}
+		t.Logf("Waiting for MySQL cluster ReadyNodes: %d\n", mysql.Status.ReadyNodes)
+		return false, nil
+	})
+	if err != nil {
+		return err
+	}
+	t.Logf("Cluster has %d ready nodes!\n", mysql.Status.ReadyNodes)
```
gabrieljackson: nit: this would be a bit clearer I think
```suggestion
	t.Logf("All MySQL cluster nodes (%d) are ready!\n", mysql.Status.ReadyNodes)
```

gabrieljackson:
@jozuenoon, after discussing the upgrade options for the new DB naming with the rest of the team, we have decided it's okay to go forward with this change as is. We don't want to introduce logic that will need to be cleaned up at a later date in order to handle the old name.

Recently, we made quite a few important changes to the operator code and tests to fix issues. Please rebase/merge master and fix any conflicts. Once tests are green we will review quickly and get this in!

Thanks!

PR #108, opened 2019-12-11
default MM to use 5.17.1 release

cpanato:
Administrative update: set default MM to use 5.17.1 release


cpanato:
will wait for https://github.com/mattermost/mattermost-operator/pull/110

PR #113, opened 2019-12-18
added TLS support during the deployment of Mattermost on Kubernetes

### LINKED ISSUE #97: Mattermost install-kubernetes TLS Ingress Controller
We have installed Mattermost with the k8s [manual](https://docs.mattermost.com/install/install-kubernetes.html). Successful! :) After that we tried to [configure](https://kubernetes.io/docs/concepts/services-networking/ingress/#tls) a TLS certificate on the Mattermost Ingress Controller, but there is no option in the `mattermost-installation.yaml`.  

``` yaml
apiVersion: networking.k8s.io/v1beta1
kind: Ingress
metadata:
  name: tls-example-ingress
spec:
  tls:
  - hosts:
    - sslexample.foo.com
    secretName: testsecret-tls
```

Will this feature come or can we solve this differently?

macevil:
#### Summary
Now it is possible to add TLS support during the deployment of Mattermost on Kubernetes. 
We tested this feature in combination with [cert-manager](https://cert-manager.io). Therefore you have to add the cert-manager issuer to your Mattermost deployment manifest. 

```yaml
apiVersion: mattermost.com/v1alpha1
kind: ClusterInstallation
metadata:
  name: your-mattermost-hostname.com
spec:
  size: 100users
  ingressName: yourmattermost.hostname.com
  ingressAnnotations:
    kubernetes.io/ingress.class: nginx
    cert-manager.io/issuer: letsencrypt-prod
  version: 5.17.1
  mattermostLicenseSecret: ""
  database:
    storageSize: 50Gi
  minio:
    storageSize: 50Gi
  elasticSearch:
    host: ""
    username: ""
    password: ""
```

The [cert-manager issuer](https://cert-manager.io/docs/tutorials/acme/ingress/#step-6-configure-let-s-encrypt-issuer) needs to be deployed into the Mattermost namespace. 

During the automatic configuration of the TLS feature a secret will be added to the cluster. The name of this secret follows this convention:

The name of the secret is identical to the ingressName but all the dots (.) are replaced by dashes (-) and the suffice "-tls-cert" will be appended. In the example above the secretName will be "yourmattermost-hostname-com-tls-cert".

As an additional point the official installation guide for Mattermost on Kubernetes needs to be updated (https://docs.mattermost.com/install/install-kubernetes.html).

#### Ticket Link
Fixes #97

### DIFF
--- a/pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
+++ b/pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
@@ -234,7 +234,7 @@ func (mattermost *ClusterInstallation) GenerateService(serviceName, selectorName
 
 // GenerateIngress returns the ingress for Mattermost
 func (mattermost *ClusterInstallation) GenerateIngress(name, ingressName string, ingressAnnotations map[string]string) *v1beta1.Ingress {
-	return &v1beta1.Ingress{
+	var ingress = &v1beta1.Ingress{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
 			Namespace: mattermost.Namespace,
@@ -269,6 +269,16 @@ func (mattermost *ClusterInstallation) GenerateIngress(name, ingressName string,
 			},
 		},
 	}
+	if ingressAnnotations["cert-manager.io/issuer"] != "" {
+
+		ingress.Spec.TLS = []v1beta1.IngressTLS{
+			{
+				Hosts:      []string{ingressName},
+				SecretName: strings.ReplaceAll(ingressName, ".", "-") + "-tls-cert",
+			},
+		}
+	}
+	return ingress
 }
 
 // GetContainerByName gets container from a deployment by name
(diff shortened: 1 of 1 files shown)


jasonblais:
@gabrieljackson A friendly reminder to help with reviews after holidays, thanks! :)

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -269,6 +269,15 @@ func (mattermost *ClusterInstallation) GenerateIngress(name, ingressName string,
 			},
 		},
 	}
+	if ingressAnnotations["cert-manager.io/issuer"] != "" {
```
cpanato: thinking a bit more, I think we should not make this hardcoded. some reasons:
- if cert-manager change the annotation name (this already happened a few months ago)
- if the user has another way to generate the TLS 

so i think we need to think a bit more, and maybe have just a flag in the config file that set the whole `v1beta1.IngressTLS` and the user need to set the annotation as well otherwise will not work.

macevil: Got it. We'll make a suggestion.

denniswinter: Besides the fact that annotation names could change in the future, there is already another annotation named cert-manager.io/cluster-issuer.

cpanato:
closing this PR due to inactivity and in favor of https://github.com/mattermost/mattermost-operator/pull/132

PR #125, opened 2020-02-10
add custom liveness and readiness probes in the spec for clusterinstallation

cpanato:
### Summary
Add option to custom define the liveness and readiness probes

### DIFF
--- a/pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
+++ b/pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
@@ -562,6 +562,8 @@ func (mattermost *ClusterInstallation) GenerateDeployment(deploymentName, ingres
 	maxUnavailable := intstr.FromInt(defaultMaxUnavailable)
 	maxSurge := intstr.FromInt(defaultMaxSurge)
 
+	liveness, readiness := setProbes(mattermost.Spec.LivenessProbe, mattermost.Spec.ReadinessProbe)
+
 	return &appsv1.Deployment{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      deploymentName,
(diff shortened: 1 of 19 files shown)


pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -758,3 +740,79 @@ func mergeEnvVars(original, new []corev1.EnvVar) []corev1.EnvVar {
...
+				Port: intstr.FromInt(8065),
+			},
+		},
+		InitialDelaySeconds: 10,
+		PeriodSeconds:       10,
+		FailureThreshold:    3,
+	}
+
+	if customLiveness.Handler != (corev1.Handler{}) {
+		liveness.Handler = customLiveness.Handler
```
gabrieljackson: Do we want users to be able to set their own handlers? I feel like being able to set things like the path and port could lead to unexpected pod behavior. I am all for being able to set things like failureThreshold, but I am not sure I see the value in being able to set the handler at this time.

Thoughts?

cpanato: no strong opinions, but imagine the user create a custom image and there they define a custom health endpoint in another path or even using a script or any other method. then will be not possible. was in this specific case i was thinking to have the entire probe object and also to help a bit in the development :)

but if you dont want that and is better to have just the other fields I'm happy to change. 

@gabrieljackson

gabrieljackson: That is something I hadn't considered, but makes a lot of sense! Thanks.

cpanato: my bad, i should improve the PR description.
mental note to myself

PR #132, opened 2020-03-13
[MM-23256] TLS support for ingress resource

cpanato:
#### Summary
This PR adds support to create TLS when creating the Ingress resource.

When user want to create the ingress and TLS support need to set the `UseIngressTLS` and the ingress annotations (for example for cert-manager) to make it work

 - Fixes: https://github.com/mattermost/mattermost-operator/issues/97

And it is related to https://github.com/mattermost/mattermost-operator/pull/113 but due inactivity we decided to move forward

#### Ticket Link
JIRA: https://mattermost.atlassian.net/browse/MM-23256

### DIFF
--- a/pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
+++ b/pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
@@ -233,8 +233,8 @@ func (mattermost *ClusterInstallation) GenerateService(serviceName, selectorName
 }
 
 // GenerateIngress returns the ingress for Mattermost
-func (mattermost *ClusterInstallation) GenerateIngress(name, ingressName string, ingressAnnotations map[string]string) *v1beta1.Ingress {
-	return &v1beta1.Ingress{
+func (mattermost *ClusterInstallation) GenerateIngress(name, ingressName string, ingressAnnotations map[string]string, useTLS bool) *v1beta1.Ingress {
+	ingress := &v1beta1.Ingress{
 		ObjectMeta: metav1.ObjectMeta{
 			Name:      name,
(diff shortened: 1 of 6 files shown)


pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -269,6 +269,17 @@ func (mattermost *ClusterInstallation) GenerateIngress(name, ingressName string,
 			},
 		},
 	}
+
+	if useTLS {
+		ingress.Spec.TLS = []v1beta1.IngressTLS{
+			{
+				Hosts:      []string{ingressName},
+				SecretName: strings.ReplaceAll(ingressName, ".", "-") + "-tls-cert",
```
gabrieljackson: Can you explain what this line is doing? Does this mean that if you want to use TLS you need a secret named `INGRESS_NAME-tls-cert` for each ingress?

If so, I think this should be documented.

cpanato: yes, this secret the `cert-manager` will populate with the certificate that will be generated. this is a standard when using TLS.
you can also create this secret manually and add the certificates in the secret and when creating the ingress you say where are your secrets.

I will document, this and other things in a followup, if is that ok for you
@gabrieljackson

cpanato: @gabrieljackson PTAL

gabrieljackson: Sounds good on a document follow-up!

PR #136, opened 2020-04-01
Mm 23681 - Adding support for custom resource labels

stylianosrigas:
This PR adds support for custom Resource labels.
At the moment the Mattermost operator is applying some default labels in all resources. With this change a user can add the resourceLabels section in the cluster installation and specify a map of labels that will be applied in all resources. 

This change covers blue/green deployments too.

A new function called ClusterInstallationSelectorLabels was created to cover selector labels. The reason is that selector labels are immutable and cannot be edited after creation.

### DIFF
--- a/deploy/crds/mattermost.com_clusterinstallations_crd.yaml
+++ b/deploy/crds/mattermost.com_clusterinstallations_crd.yaml
@@ -1186,6 +1186,10 @@ spec:
                 set by 'Size'.
               format: int32
               type: integer
+            resourceLabels:
+              additionalProperties:
+                type: string
+              type: object
             resources:
               description: Defines the resource requests and limits for the Mattermost
                 app server pods.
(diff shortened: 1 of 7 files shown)


stylianosrigas:
> Hmm, I was thinking that we may want to be able to set different resource labels for normal, blue, and green deployments instead of all reusing the same custom labels. Thoughts?

0/5 on that. If we want to do this we need to create separate objects I guess, like blueResourceLabels and greenResourceLabels. @cpanato @jwilander any thoughts?

cpanato:
yes, we will need to create different types
but if the label selectors work with the ones the operator i think we are good for now

jwilander:
I think I'd want the ability to add a label to all the resources that the operator creates but then also the ability to add different labels to blue/green independently. 0/5 on that

gabrieljackson:
Yeah, I don't think it is a hard requirement either right now, but I do think it makes sense to have different label options for Blue/Green in the long run. Considering Blue, Green,and Canary all use `AppDeployment` under the hood, we should just be able to add one more label value there and support all of these deployment types: https://github.com/mattermost/mattermost-operator/pull/136/files#diff-56ddcfa304b2a27f39c0c8d0ccb50786R117

stylianosrigas:
Cool thanks for the ideas guys. I will work on that to support separate labels for blue green as well. I think it makes sense to introduce this with this PR ;)

PR #137, opened 2020-04-01
Bump internal version variable

gabrieljackson:
Just to match our latest release tag.

### DIFF
--- a/version/version.go
+++ b/version/version.go
@@ -4,7 +4,7 @@ import (
 	"fmt"
 )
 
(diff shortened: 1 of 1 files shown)


version/version.go
```diff
@@ -4,7 +4,7 @@ import (
 	"fmt"
 )
 
-var version = "1.1.1"
+var version = "1.4.0"
```
cpanato, line 7: humm, forgot that

PR #144, opened 2020-05-16
Remove hard-coded cluster DNS for minio

### LINKED ISSUE #143: Minio InitContainer fails with custom cluster DNS
https://github.com/mattermost/mattermost-operator/blob/018b4f3988a7a9a6e311038696872093e80412ac/pkg/controller/clusterinstallation/minio.go#L129

`"%s.%s.svc.cluster.local:%d"` is used for the minio connection string.

I have a custom DNS for my cluster so hard-coding the `cluster.local` in there makes it unable start for me.

mterhar:
#### Summary

Allows the operator to work with clusters that don't use the default "cluster.local" domain name.  SQL's reference uses {service name}.{namespace} without the `svc.cluster.local` for its connection string. 

This change applies that same DNS approach to the minio service. 

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-operator/issues/143

#### Release Note

```release-note
Fixes bug in minio connection when not using default "cluster.local" cluster domain. 
```

### DIFF
--- a/pkg/controller/clusterinstallation/minio.go
+++ b/pkg/controller/clusterinstallation/minio.go
@@ -126,6 +126,6 @@ func (r *ReconcileClusterInstallation) getMinioService(mattermost *mattermostv1a
 		return "", err
 	}
 
-	connectionString := fmt.Sprintf("%s.%s.svc.cluster.local:%d", minioService.Name, mattermost.Namespace, minioService.Spec.Ports[0].Port)
+	connectionString := fmt.Sprintf("%s.%s:%d", minioService.Name, mattermost.Namespace, minioService.Spec.Ports[0].Port)
 	return connectionString, nil
 }


mterhar:
I forgot to hit submit on the CLA so I didn't sign it until after the check failed.  Not sure how to re-run that check without pushing a pointless commit.

gabrieljackson:
Thanks for the PR @mterhar! I am going to do some testing with this to ensure it works in some different environments we have and will get back with you.

gabrieljackson:
Thanks @mterhar!

mterhar:
Is there a timeframe for how long this will take to show up in the "released" containers? 

I just pulled from `https://raw.githubusercontent.com/mattermost/mattermost-operator/master/docs/mattermost-operator/mattermost-operator.yaml` and it looks like the operator still has the hard-coded DNS.

cpanato:
@mterhar Hello, we are planning a new release for early next month, but no guarantees yet. We need to finish some other things.

In the meantime, if you want you can use `latest`

PR #145, opened 2020-05-25
MM-25480: added ability to create Mattermost containers with CA certificate files

gsagula:
#### Summary
This PR enhances the operator so that it can retrieve CA certificate PEMs from a secret and mount them as files inside a Mattermost container. Certificates files are passed via a secret named `ca-certificate-files` which holds key/value pairs of a filename and a certificate PEM correspondently. The files are mounted inside the Mattermost container at `/etc/ssl/certs`. This feature is needed in situations which the container needs to authenticate with other systems. For example, establish a secure connection with a RDS database instance. 

#### Ticket Link
Fixes: https://mattermost.atlassian.net/browse/MM-25480
Unblock: https://github.com/mattermost/mattermost-cloud/pull/221

#### Release Note
```release-note
Adds ability to retrieve CA certificate PEMs via K8s cluster secret and mount them inside a Mattermost container's file system. Certificates are passed via a secret named `ca-certificate-files` which holds key/value pairs of a filename and a certificate PEM correspondently. The files are then mounted inside the Mattermost's container at `/etc/ssl/certs`.
```


pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -53,6 +53,16 @@ const (
...
 	DefaultMaxFileSize = 1000
 
+	// MattermostAppContainerName is the name of the container which runs Mattermost application
+	MattermostAppContainerName = "mattermost"
+
+	// CertificateFilesSecretName is the name of a secret used for mounting CA certificate files in the
+	// the container that runs the Mattermos application.
+	CertificateFilesSecretName = "ca-certificate-files"
+	// CertificateFilesPath is the path where the certificate files will be mounted. Note that this path
+	// is valid for Debian and Ubuntu but may not work for other distributions.
```
cpanato, line 63: we use Alpine as base image.

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -514,23 +521,41 @@ func (mattermost *ClusterInstallation) GenerateDeployment(deploymentName, ingres
...
 	podAnnotations := map[string]string{}
+
+	// CA Certificate Files
+	volumesMount = append(volumesMount, corev1.VolumeMount{
+		Name:      CertificateFilesSecretName,
+		MountPath: CertificateFilesPath,
+		ReadOnly:  true,
+	})
+
+	volumes = append(volumes, corev1.Volume{
```
cpanato, line 536: this will be mounted only if we have a license, is this the behavior we want?

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -514,23 +521,41 @@ func (mattermost *ClusterInstallation) GenerateDeployment(deploymentName, ingres
...
+		Name: CertificateFilesSecretName,
+		VolumeSource: corev1.VolumeSource{
+			Secret: &corev1.SecretVolumeSource{
+				SecretName: CertificateFilesSecretName,
```
cpanato, line 540: if the secret does not exist this will fail and the pod will not start. We need to validate if the secret exist and have the correct format

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
gsagula: It should be the same for Alpine. Are the container images used in dev also built from this distribution?

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
cpanato: Also if you mount this volume in the `/etc/ssl/certs` if going to overwrite all the other certificates we install when we build the Mattermost image, [see](https://github.com/mattermost/mattermost-server/blob/master/build/Dockerfile#L12)

to avoid this you will need to mount like this
```
    volumeMounts:
    - name: foo
      mountPath: /etc/ssl/certs/rds-ca-2019-root.pem
      subPath: rds-ca-2019-root.pem
      readOnly: true
```

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
cpanato: everything and alpine have the same path, just checked

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
gsagula: Right, that's what I had before. I do the change!

cpanato: the `CertificateFilesSecretName` should be configurable as well and if not set it can use the default value

gsagula: I was thinking about using this constant in `mattermost-cloud` but if we are going to make this configurable, it changes everything.

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
gsagula: That was the intention, yes. This way we don't even need to bother by checking if we have licenses.

gsagula:
@cpanato Thanks for the review. PTAL

One extra modification: In addition to make the secret name configurable, I also provided the same option for [path](https://github.com/mattermost/mattermost-operator/pull/145/files#diff-56ddcfa304b2a27f39c0c8d0ccb50786R105) since it may vary depending on the distro.  Both fields are optional and with correspondent default values when empty.

cpanato:
@gsagula can you squash the commits?

pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -387,3 +409,66 @@ func (r *ReconcileClusterInstallation) fetchRunningUpdateJob(mi *mattermostv1alp
...
 	return job, err
 }
+
+// getCertificateFilesSecret returns a secret with certificate files or error
+func (r *ReconcileClusterInstallation) getCertificateFilesSecret(mattermost *mattermostv1alpha1.ClusterInstallation) (*corev1.Secret, error) {
+	certificatesObjKey := types.NamespacedName{
+		Namespace: mattermost.Namespace,
+	}
+
+	if mattermost.Spec.Certificates.SecretName == "" {
```
cpanato, line 419: here maybe just
```suggestion
        certificatesObjKey.Name = v1alpha1.DefaultCertificatesSecretName
	if mattermost.Spec.Certificates.SecretName != "" {
		certificatesObjKey.Name = mattermost.Spec.Certificates.SecretName
	}
```

pkg/apis/mattermost/v1alpha1/clusterinstallation_types.go
```diff
@@ -85,6 +85,24 @@ type ClusterInstallationSpec struct {
 	// Defines the probe to check if the application is ready to accept traffic.
 	// +optional
 	ReadinessProbe corev1.Probe `json:"readinessProbe,omitempty"`
+
+	// Certificates tells the operator which secret contains certificate PEM file(s).
+	// Each file will be mounted as volume inside the Mattermost container.
+	// +optional
+	Certificates Certificates `json:"certificates,omitempty"`
```
cpanato, line 92: maybe is worth explicitly say which type of certificate, in this case i would name it `CACertificates`

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -851,3 +868,39 @@ func setProbes(customLiveness corev1.Probe, customReadiness corev1.Probe) (*core
...
 }
+
+func (mattermost *ClusterInstallation) getCertificateVolume(volumename, filename string) (*corev1.VolumeMount, *corev1.Volume) {
+	volumeMount := corev1.VolumeMount{
+		Name:     volumename,
+		SubPath:  filename,
+		ReadOnly: true,
+	}
+
+	if mattermost.Spec.Certificates.Path == "" {
```
cpanato, line 879: same applies here

pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -851,3 +868,39 @@ func setProbes(customLiveness corev1.Probe, customReadiness corev1.Probe) (*core
...
+					{
+						Key:  filename,
+						Path: filename,
+					},
+				},
+			},
+		},
+	}
+
+	if mattermost.Spec.Certificates.SecretName == "" {
```
cpanato, line 899: and here

gsagula:
> @gsagula can you squash the commits?

@cpanato Squash will destroy the commit's history. Do you want me to do that right now?

cpanato:
i prefer to have just one commit that says what that is doing instead of a million of commits, if you want to keep is fine, i will squash in the merge then

gsagula:
> i prefer to have just one commit that says what that is doing instead of a million of commits, if you want to keep is fine, i will squash in the merge then

Sounds good, I do it now if you prefer :)

cpanato:
go for it!

gsagula:
@cpanato @gabrieljackson Friendly reminder that the changes have been submitted and the PR is ready for another pass. Thanks!

jwilander:
Per discussion in cloud team channel we're going to pause on this for now.

PR #154, opened 2020-06-19
update checker: remove probes from the job

cpanato:
#### Summary
Removing the probes from the Job (update mattermost checker) because this is a short live execution and we don't need to have the readiness probe to start receiving traffic because this will not serve any request.

Also, the liveness can kill the job before it completes its execution, which is the case in community-daily (more context [see this 🧵 ](https://community-daily.mattermost.com/core/pl/1it9rb319pgs3x33j4wsyh4f3h) 

The job will run a specific command and will exit with success or fail which is we will check later. If the image does not exist it will fail in the initialization part and we don't need the probes for that.

#### Ticket Link
n/a

#### Release Note

```release-note
job: remove probes from the job image checker
```

### DIFF
--- a/pkg/controller/clusterinstallation/mattermost.go
+++ b/pkg/controller/clusterinstallation/mattermost.go
@@ -226,6 +226,13 @@ func (r *ReconcileClusterInstallation) launchUpdateJob(
 		},
 	}
 
+	// We dont need to validate the readiness/liveness for this short live job
+	// if the job fails it will get another later.
+	for i := range job.Spec.Template.Spec.Containers {
+		job.Spec.Template.Spec.Containers[i].LivenessProbe = nil
+		job.Spec.Template.Spec.Containers[i].ReadinessProbe = nil
+	}
+
 	// Override values for job-specific behavior.
 	job.Spec.Template.Spec.RestartPolicy = corev1.RestartPolicyNever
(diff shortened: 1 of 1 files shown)


pkg/controller/clusterinstallation/mattermost.go
```diff
@@ -226,6 +226,11 @@ func (r *ReconcileClusterInstallation) launchUpdateJob(
 		},
 	}
 
+	// We dont need to validate the readiness/liveness for this short live job
+	// if the job fails it will get another later.
+	job.Spec.Template.Spec.Containers[0].LivenessProbe = nil
+	job.Spec.Template.Spec.Containers[0].ReadinessProbe = nil
```
gabrieljackson, line 232: This is a little dangerous as we are referencing `Containers[0]` without checking the length of the slice. It should be fine because we should always have at least one container, but maybe it makes sense to do something like this instead?

```
for i := range job.Spec.Template.Spec.Containers {
  	job.Spec.Template.Spec.Containers[i].LivenessProbe = nil
	job.Spec.Template.Spec.Containers[i].ReadinessProbe = nil
}
```

cpanato: makes sense

cpanato: done @gabrieljackson PTAL

PR #156, opened 2020-07-03
Add external database init check

gabrieljackson:
When using an external database, the Mattermost app pods previously
did not check if the database connection was available before
starting the app containers. This change adds this database check
as an option that can be used by specifying a new key and value in
the database secret.

To accomplish this, the database flow of the controller was refactored.
Also, logging and error message improvements were made across the
operator.

This option will be documented in a follow-up PR once we have been
able to test it internally.

Fixes https://mattermost.atlassian.net/browse/MM-26695

```release-note
Add external database init check
```


pkg/apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -324,17 +325,42 @@ func (mattermost *ClusterInstallation) GenerateDeployment(deploymentName, ingres
...
+					},
+				},
+				Command: []string{
+					"sh", "-c",
+					"until curl --max-time 5 $DB_CONNECTION_CHECK_URL; do echo waiting for mysql; sleep 5; done;",
+				},
+			})
+		}
 	} else {
 		masterDBEnvVar.Value = fmt.Sprintf(
```
cpanato, line 354: in the future, we should move this to a secret and inject that as env var

gabrieljackson: Yep, I already have a ticket for this: https://mattermost.atlassian.net/browse/MM-26742

gabrieljackson:
Yeah, we definitely want to document this, but as I mentioned in my PR description. I would like to do it in a follow-up PR once we test it some internally.

PR #160, opened 2020-07-13
probes: add startup probe to take care for the starting up process

cpanato:
#### Summary
Sometimes, we need to deal the MM require an additional startup time on their first initialization. 
In such cases, it can be tricky to set up liveness probe parameters without compromising the fast response to deadlocks that motivated such a probe. 
The trick is to set up a startup probe with the same command, HTTP or TCP check, with a failureThreshold * periodSeconds long enough to cover the worse case startup time.

So after some knowledge refresh I discovery the new probe called `startupProbe` that works only in the initialization part, and after that the `livenessProbe` assume.

#### Ticket Link
N/a

#### Release Note

```release-note
probes: add startup probe to take care for the starting up process
```

/cc @jwilander

### DIFF
--- a/pkg/mattermost/helpers.go
+++ b/pkg/mattermost/helpers.go
@@ -44,7 +44,7 @@ func mergeEnvVars(original, new []corev1.EnvVar) []corev1.EnvVar {
 	return original
 }
 
-func setProbes(customLiveness corev1.Probe, customReadiness corev1.Probe) (*corev1.Probe, *corev1.Probe) {
+func setProbes(customLiveness, customStartup, customReadiness corev1.Probe) (*corev1.Probe, *corev1.Probe, *corev1.Probe) {
 	liveness := &corev1.Probe{
 		Handler: corev1.Handler{
 			HTTPGet: &corev1.HTTPGetAction{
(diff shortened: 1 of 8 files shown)


go.mod
```diff
@@ -28,5 +28,6 @@ require (
 
 replace (
 	github.com/Azure/go-autorest => github.com/Azure/go-autorest v13.3.2+incompatible // Required by OLM
+	golang.org/x/tools => golang.org/x/tools v0.0.0-20200713011307-fd294ab11aed
```
gabrieljackson, line 31: Ah, that explains why I was still seeing vendor issues in master.

cpanato: figure out today, the lib is updated almost daily so we need to pin until we figure out how to fix that and not have as hard dependency

pkg/mattermost/helpers.go
```diff
@@ -81,6 +81,42 @@ func setProbes(customLiveness corev1.Probe, customReadiness corev1.Probe) (*core
...
+		startUp.FailureThreshold = customStartup.FailureThreshold
+	}
+
+	if customStartup.SuccessThreshold != 0 {
+		startUp.SuccessThreshold = customStartup.SuccessThreshold
+	}
+
+	if customStartup.FailureThreshold != 0 {
+		startUp.FailureThreshold = customStartup.FailureThreshold
+	}
```
gabrieljackson, line 118: This is the second time we check and set `customStartup.FailureThreshold` in this function. Did you mean to check something else or is this just a duplicate?

cpanato: was a copy and paster, the other have the same issue :)

cpanato: will wait for [this](https://github.com/mattermost/mattermost-operator/pull/162) to be merged to update this PR

PR #161, opened 2020-07-13
Use database secret values with MySQL operator connections

gabrieljackson:
This changes how the database username and password are injected
into the Mattermost app pod when using the MySQL operator. These
values are now directly referenced from the database secret as
env vars.

Fixes https://mattermost.atlassian.net/browse/MM-26742

```release-note
Use database secret values with MySQL operator connections
```


pkg/mattermost/mattermost.go
```diff
@@ -152,32 +151,61 @@ func GenerateDeployment(mattermost *mattermostv1alpha1.ClusterInstallation, dbIn
...
+							Name: dbInfo.SecretName,
+						},
+						Key: "PASSWORD",
+					},
+				},
+			},
+			{
+				Name: "MM_SQLSETTINGS_DATASOURCEREPLICAS",
+				Value: fmt.Sprintf(
+					"$(MYSQL_USERNAME):$(MYSQL_PASSWORD)@tcp(%s-mysql.%s:3306)/%s?readTimeout=30s&writeTimeout=30s",
```
cpanato, line 192: maybe we can add this in the secret and then avoid to make the data public available in the deployment. what do you think?

cpanato: i will explain to you in a call, but lgtm as is

PR #164, opened 2020-07-21
add option to use digest id instead of tag

cpanato:
#### Summary
Add the option to use digest instead of tags.
Some users use digest instead of tags because that digests are immutable and tags are mutable.

In our cloud use case:
We pull the master tag daily to update community-daily and the process of that is:
- pull the master tag
- retag master adding a suffix, ie ci_build env var
- push the newly tagged image which is the same as the current master
- update the manifest to use the new image

using the digest we avoid the tagging and pushing a duplicate image

#### Ticket Link
n/a

#### Release Note

```release-note
add option to use digest id instead of tag
```

### DIFF
--- a/pkg/apis/mattermost/v1alpha1/clusterinstallation_test.go
+++ b/pkg/apis/mattermost/v1alpha1/clusterinstallation_test.go
@@ -1,6 +1,7 @@
 package v1alpha1
 
 import (
+	"fmt"
 	"testing"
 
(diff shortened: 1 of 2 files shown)


cpanato:
if want to read more about:
- https://renovate.whitesourcesoftware.com/blog/overcoming-dockers-mutable-image-tags/
- https://success.docker.com/article/images-tagging-vs-digests

PR #170, opened 2020-08-03
resources: update cpu/mem requests and limits

cpanato:
#### Summary
Update the CPU and memory requests and limits for the services that we deploy.

Some requests were set to a huge value and maybe we don't need that, since the request is the minimum needed to the service start running and looking at our community server the CPU was not requesting too much.

In one type of installation (5000 users ), it was requesting 3.5 CPU this is to a huge consumption CPU app, out services is not that hungry for CPU, and with that in a "normal" cluster will be never scheduled and most likely will run alone in one node, and will need more nodes to host the side apps (db/minio) because those were requesting a lot of CPU as well.

This PR reduces the requests for bot CPU/memory and try to standardize those.

If we merge that we might want to merge in the `release-1.6` branch and release `v1.6.2` as well

#### Release Note

```release-note
resources: update cpu/mem requests and limits
```


cpanato:
I will cherry-pick this to the release-1.6 branch

PR #171, opened 2020-08-11
Add new env var for the license

cpanato:
#### Summary
this new env var is only available for release > 5.26, for older releases that make no difference if we set.

this sets the `MM_LICENSE` to be used to hold the license information

#### Ticket Link
Jira: https://mattermost.atlassian.net/browse/MM-27649

#### Release Note

```release-note
Add new env var for license for MM releases > 5.26
```

### DIFF
--- a/pkg/mattermost/helpers.go
+++ b/pkg/mattermost/helpers.go
@@ -1,6 +1,11 @@
 package mattermost
 
 import (
+	"regexp"
+
+	"github.com/blang/semver/v4"
+	"github.com/pkg/errors"
+
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/util/intstr"
 )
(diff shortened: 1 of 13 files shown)


pkg/mattermost/helpers.go
```diff
@@ -143,3 +148,25 @@ func setProbes(customLiveness, customStartup, customReadiness corev1.Probe) (*co
 
 	return liveness, startUp, readiness
 }
+
+func checkMattermostImageVersion(imageName, version string) error {
+	var re = regexp.MustCompile(`mattermost/mattermost-.*[team|enterprise]-edition$`)
+	if !re.MatchString(imageName) {
+		return errors.Errorf("not using the mattermost official images so cannot validate the version")
```
gigawhitlocks, line 155: Will this make it impossible to use the Operator with a custom-built image?

gigawhitlocks: I guess the `.*` in the middle of the pattern will accept anything so someone could always put a custom tag name in that part of the string.

pkg/mattermost/helpers.go
```diff
@@ -143,3 +148,25 @@ func setProbes(customLiveness, customStartup, customReadiness corev1.Probe) (*co
...
+	}
+
+	v, err := semver.Parse(version)
+	if err != nil {
+		return errors.Wrap(err, "failed to parse the version, maybe it is using ")
```
gabrieljackson, line 160: `maybe it is using `? I am left wondering what it could be using. :D

pkg/mattermost/helpers.go
gabrieljackson: This is a good point. If someone wants to make their own custom mattermost image I would think we would want them to still be able to use our operator. Either way:
```suggestion
		return errors.Error("not using the mattermost official images so cannot validate the version")
```

cpanato:
i will close this for now to implement a pod checker version

PR #175, opened 2020-09-12
Refactor ClusterInstallation state logic

gabrieljackson:
This change refactors all controller logic related to updating the
state field of a ClusterInstallation's status. This was done to
correct an issue where the state value could be temporarily
incorrect. As part of this refactor, the following changes were
also made:

 - Improved logging of status changes as well as other important
   events. More logging fields added as well.
 - Improved error handling by returning wrapped errors and letting
   the caller log as necessary.
 - Removed false logs that showed up when not using the mysql or
   minio operators.
 - Cherry-pick a change to remove the startup probe being set in
   Mattermost deployments.

Fixes https://mattermost.atlassian.net/browse/MM-28636

```release-note
Refactor ClusterInstallation state logic
```

### DIFF
--- a/pkg/controller/clusterinstallation/controller.go
+++ b/pkg/controller/clusterinstallation/controller.go
@@ -36,7 +36,7 @@ func Add(mgr manager.Manager) error {
 
 // newReconciler returns a new reconcile.Reconciler
 func newReconciler(mgr manager.Manager) reconcile.Reconciler {
-	return &ReconcileClusterInstallation{client: mgr.GetClient(), scheme: mgr.GetScheme(), state: mattermostv1alpha1.Reconciling}
+	return &ReconcileClusterInstallation{client: mgr.GetClient(), scheme: mgr.GetScheme()}
 }
 
 // add adds a new Controller to mgr with r as the reconcile.Reconciler
@@ -96,15 +96,6 @@ type ReconcileClusterInstallation struct {
 	// that reads objects from the cache and writes to the apiserver.
 	client client.Client
 	scheme *runtime.Scheme
-	state  mattermostv1alpha1.RunningState
-}
-
-func (r *ReconcileClusterInstallation) setReconciling() {
-	r.state = mattermostv1alpha1.Reconciling
-}
-
-func (r *ReconcileClusterInstallation) setStable() {
-	r.state = mattermostv1alpha1.Stable
 }
 
 // Reconcile reads the state of the cluster for a ClusterInstallation object and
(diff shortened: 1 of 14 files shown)


pkg/controller/clusterinstallation/controller.go
```diff
@@ -118,27 +109,23 @@ func (r *ReconcileClusterInstallation) Reconcile(request reconcile.Request) (rec
...
 		return reconcile.Result{}, err
 	}
 
-	if mattermost.Status.State != r.state {
-		status := mattermost.Status
-		status.State = r.state
-		err = r.updateStatus(mattermost, status, reqLogger)
+	// Set a new ClusterInstallation's state to reconciling.
+	if len(mattermost.Status.State) == 0 {
+		err = r.setStateReconciling(mattermost, reqLogger)
```
jwilander, line 127: Is this the main fix?

jwilander (commented):
Looks good - just want to understand the primary fix for the issue

pkg/controller/clusterinstallation/controller.go
gabrieljackson: The main part of the fix is that fact that we no longer keep track of the state as an in-memory var on the reconciler. Instead, we just compare the previous state variable to whatever the reconciler thinks the state should be in that moment. The in-memory state var was how new CIs had the possibility of being set to stable right away. I am surprised it didn't also cause other issues.

This part specifically is just the reconciler setting the state on a new CI that was just created. It should only happen once per custom resource that is created.

jwilander: Cool - got it 👍

jwilander (approved):

pkg/controller/clusterinstallation/controller.go
```diff
@@ -118,27 +109,23 @@ func (r *ReconcileClusterInstallation) Reconcile(request reconcile.Request) (rec
...
 		if err != nil {
-			r.setReconciling()
 			return reconcile.Result{}, err
 		}
```
Szymongib, line 130: Just one question: Is there a reason not to log the error here?
I think it was previously logged inside `updateStatus` but as it is no longer the case, you might have missed it.

Szymongib (commented):
👍

Szymongib (approved):

stylianosrigas (approved):
Thanks Gabe, LGTM :)

pkg/controller/clusterinstallation/controller.go
gabrieljackson: Good question! The returned errors from the reconcile function are always logged so we just let that handle it. In this case the error is that the status could not be set, but in other places we need to log that error separately.

PR #183, opened 2020-10-02
DOPS-216: Add MM_INSTALL_TYPE=kubernetes

metanerd:
#### Summary
 Sets an environment variable to get an idea of the preferred installation type. 

#### Ticket Link

https://mattermost.atlassian.net/jira/software/projects/DOPS/boards/66/backlog?selectedIssue=DOPS-216

#### Release Note

```release-note
NONE
```

### DIFF
--- a/pkg/mattermost/mattermost.go
+++ b/pkg/mattermost/mattermost.go
@@ -373,7 +373,7 @@ func GenerateDeployment(mattermost *mattermostv1alpha1.ClusterInstallation, dbIn
 		},
 		{
 			Name:  "MM_INSTALL_TYPE",
-			Value: "kubernetes",
+			Value: "kubernetes-operator",
 		},
 	}
 
--- a/pkg/mattermost/mattermost.go
+++ b/pkg/mattermost/mattermost.go
@@ -371,6 +371,10 @@ func GenerateDeployment(mattermost *mattermostv1alpha1.ClusterInstallation, dbIn
 			Name:  "MM_CLUSTERSETTINGS_CLUSTERNAME",
 			Value: "production",
 		},
+		{
+			Name:  "MM_INSTALL_TYPE",
+			Value: "operator",
+		},
 	}
 
 	valueSize := strconv.Itoa(defaultMaxFileSize * sizeMB)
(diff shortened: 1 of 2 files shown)


pkg/mattermost/mattermost.go
```diff
@@ -371,6 +371,10 @@ func GenerateDeployment(mattermost *mattermostv1alpha1.ClusterInstallation, dbIn
 			Name:  "MM_CLUSTERSETTINGS_CLUSTERNAME",
 			Value: "production",
 		},
+		{
+			Name:  "MM_INSTALL_TYPE",
+			Value: "operator",
```
metanerd, line 376:
```suggestion
			Value: "kubernetes",
```

gabrieljackson (approved):
Looks good!

I did want to check that we want to use `kubernetes` as the value for this though. We have at least two types of kubernetes deployments with helm and the operator. Do we want to be more specific with something like `kubernetes-operator` or something similar?

metanerd:
@grundleborg ^

grundleborg:
@gabrieljackson @metanerd making it more specific in that case would be great - you can choose the term (so `kubernetes_operator` is great).

cpanato (changes requested):
@metanerd can you update the variable name per George request?
otherwise lgtm

pkg/mattermost/mattermost.go
```diff
@@ -371,6 +371,10 @@ func GenerateDeployment(mattermost *mattermostv1alpha1.ClusterInstallation, dbIn
 			Name:  "MM_CLUSTERSETTINGS_CLUSTERNAME",
 			Value: "production",
 		},
+		{
+			Name:  "MM_INSTALL_TYPE",
+			Value: "kubernetes",
```
metanerd, line 376:
```suggestion
			Value: "kubernetes-operator",
```

cpanato (approved):

cpanato:
adding @grundleborg as a reviewer as well since he asks for the change

grundleborg (approved):
LGTM

PR #184, opened 2020-10-06
Add MM_LICENSE variable

cpanato:
#### Summary
Add MM_LICENSE variable

when the mattermost version is greater then 5.28 we use the can set the MM_LICENSE instead of the old env var for the license, this makes the recycle of the license easier and doesn't need any user intervention. Because using the old variable MM store the license in the DB and when you replace the license it does not update the license.

To check the MM version we create a sleep pod using the image and run the CLI command to get the version. The pod will be up and running during the reconciliation phase and will be terminated when it completes.
If for some reason it cannot get the version it will assume that is an older version. The command we run only works for 5.28 or greater so if the command fails to run that means it is not a >= 5.28 version.

thanks for @Szymongib for helping here as well

#### Ticket Link
JIRA: https://mattermost.atlassian.net/browse/MM-27649

#### Release Note

```release-note
Add MM_LICENSE variable
```


pkg/controller/clusterinstallation/client.go
```diff
@@ -0,0 +1,13 @@
+package clusterinstallation
+
+import (
+	"k8s.io/client-go/kubernetes"
+	"k8s.io/client-go/rest"
+)
+
+// GenerateK8sClient create client for kubernetes
+func GenerateK8sClient() (*kubernetes.Clientset, *rest.Config) {
```
Szymongib, line 9: I think this function is not used. Can we remove it?

pkg/controller/clusterinstallation/controller.go
```diff
@@ -94,8 +125,11 @@ var _ reconcile.Reconciler = &ReconcileClusterInstallation{}
 type ReconcileClusterInstallation struct {
 	// This client, initialized using mgr.Client() above, is a split client
 	// that reads objects from the cache and writes to the apiserver.
-	client client.Client
-	scheme *runtime.Scheme
+	client      client.Client
+	config      *rest.Config
+	podExecutor PodExecutor
+	scheme      *runtime.Scheme
+	state       mattermostv1alpha1.RunningState
```
Szymongib, line 132: I think `config` is not used anymore, also I believe `state` was removed by @gabrieljackson and is no longer used either.

pkg/controller/clusterinstallation/mattermost_test.go
```diff
@@ -47,10 +53,19 @@ func TestCheckMattermost(t *testing.T) {
...
 	apis.AddToScheme(scheme.Scheme)
 	s := scheme.Scheme
 	s.AddKnownTypes(mattermostv1alpha1.SchemeGroupVersion, ci)
-	r := &ReconcileClusterInstallation{client: fake.NewFakeClient(), scheme: s}
+
+	testServer, _, _ := testServerEnv(t, 200)
+	defer testServer.Close()
+
+	config := restConfig(testServer)
+	//cs := kFake.NewSimpleClientset()
```
Szymongib, line 66: I think this can be also cleaned up, as the `PodExecutor` is mocked and we do not call anything actually.

pkg/controller/clusterinstallation/mattermost_test.go
```diff
@@ -228,11 +260,20 @@ func TestCheckMattermostExternalDB(t *testing.T) {
...
 	s := scheme.Scheme
 	s.AddKnownTypes(mattermostv1alpha1.SchemeGroupVersion, ci)
-	r := &ReconcileClusterInstallation{client: fake.NewFakeClient(), scheme: s}
+
+	testServer, _, _ := testServerEnv(t, 200)
+	defer testServer.Close()
+
+	c := restConfig(testServer)
+	//rc, err := kubernetes.NewForConfig(c)
+	//require.NoError(t, err)
```
Szymongib, line 274: Same as above, we should be able to just remove it.

pkg/controller/clusterinstallation/mattermost_test.go
```diff
@@ -353,3 +394,33 @@ func TestCheckMattermostExternalDB(t *testing.T) {
...
+		Username: "user",
+		Password: "pass",
+	}
+}
+
+// func restClient(testServer *httptest.Server) (*RESTClient, error) {
+func restClient(config *rest.Config) *rest.RESTClient {
+	c, _ := rest.RESTClientFor(config)
+	return c
+}
```
Szymongib, line 426: I think we will not need those functions anymore.

Szymongib (commented):
Looks great!
I wonder if we could go for a simpler solution in the future like for example introducing `version` endpoint to the Mattermost server and just call it?

gabrieljackson:
I will review after merge conflicts have been addressed.

gabrieljackson (commented):
Just leaving a comment to remove this from my review list. Please re-request a review when ready.

cpanato:
I think this might need to rethink for v1beta1, so I will close this, and @gabrieljackson and @Szymongib can add this for v1beta1 planning

PR #189, opened 2020-10-28
MM-29717 - Add ability to limit number of installation in Reconciling state

Szymongib:
#### Summary

To prevent the Operator from starting to roll a large number of Pods at the same time, we need a mechanism to limit the number of ClusterInstallations that are being worked on at the same time.
PR introduces the following changes:
- Add environment variables to control limits and requeuing delay.
    - `MAX_RECONCILING_INSTALLATIONS` - indicates the maximum number of Installations in the `Reconciling` state.
    - `REQUEUE_ON_LIMIT_DELAY` - indicates delay with which Installations will be requeued if the limit is reached.
- Improve health check.
- Use non-cached API client in places where the latest data is needed.
- Fix logger used by the Operator.

The health check improvement now verifies that:
- The Deployment was picked up by the Deployment Controller.
- The ReplicaSet was created and observed by the Controller.

Previously the race condition might occur when the old Pods did not start to terminate and new ones were not yet created. In this situation, the Operator would decide that Installation is healthy and set into a `Stable` state, however, the invalid state would not last for a long time as the next reconciliation would pick up the unhealthy Pods and fix the state. This is not a case if we limit the number of Installations in the `Reconciling` state as the Installations in the invalid state cannot be fixed in case the limit was reached. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-29717

#### Release Note

```release-note
Limit number of Cluster Installations in the 'Reconciling' state.
```


controllers/mattermost/clusterinstallation/controller.go
```diff
@@ -66,6 +69,20 @@ func (r *ClusterInstallationReconciler) Reconcile(request ctrl.Request) (ctrl.Re
 		return reconcile.Result{}, err
 	}
 
+	if mattermost.Status.State != mattermostv1alpha1.Reconciling {
+		var clusterInstallations mattermostv1alpha1.ClusterInstallationList
+		err = r.Client.List(context.TODO(), &clusterInstallations)
+		if err != nil {
+			return reconcile.Result{}, errors.Wrap(err, "Failed to list clusterInstallations")
```
gabrieljackson, line 76: nit:
```suggestion
			return reconcile.Result{}, errors.Wrap(err, "failed to list ClusterInstallations")
```

gabrieljackson (approved):
Awesome changes! The new config options for controlling rollout are very flexible. Also, thanks for fixing the logger issue. ;)

stylianosrigas (approved):
LGTM

PR #192, opened 2020-10-30
Do not return error if health check fails

Szymongib:
#### Summary

To prevent the exponential backoff when Cluster Installation health check fails, do not return reconciliation error but requeue the request with a constant delay.

#### Release Note

```release-note
Do not return reconciliation error if the health check fails.
```

### DIFF
--- a/controllers/mattermost/clusterinstallation/controller.go
+++ b/controllers/mattermost/clusterinstallation/controller.go
@@ -22,6 +22,8 @@ import (
 	"github.com/mattermost/mattermost-operator/pkg/database"
 )
 
+const healthCheckRequeueDelay = 6 * time.Second
+
 // ClusterInstallationReconciler reconciles a ClusterInstallation object
 type ClusterInstallationReconciler struct {
 	client.Client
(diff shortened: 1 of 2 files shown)


controllers/mattermost/clusterinstallation/controller.go
```diff
@@ -135,7 +137,8 @@ func (r *ClusterInstallationReconciler) Reconcile(request ctrl.Request) (ctrl.Re
 	if err != nil {
 		r.setStateReconcilingAndLogError(mattermost, reqLogger)
 		r.updateStatus(mattermost, status, reqLogger)
-		return reconcile.Result{RequeueAfter: time.Second * 3}, err
+		reqLogger.Error(err, "Error checking ClusterInstallation health")
```
jwilander, line 140: Do we even want to log an error here? Might be better to just log it as `info` since it's not an actionable error it's just that the reconciliation hasn't completed, right?

Szymongib: I used `Error` to basically be aligned with what was logged before and technically the health check function can return an error when it, for example, cannot list some resources. Either way, I am open to change it - 0/5 on that.

PR #194, opened 2020-11-09
Add Job that initializes MM database

Szymongib:
#### Summary

To avoid a race condition when two Mattermost pods try to initialize the DB at the same time it should be done before pods startup.

Changes in PR:
- Add a Job that initializes the Mattermost database
- Add init container to Mattermost pod to wait for setup Job to finish
- Adjust RBAC of Mattermost and Operator

#### Release Note

```release-note
Add Job that initializes Mattermost database
```

### DIFF
--- a/controllers/mattermost/clusterinstallation/mattermost.go
+++ b/controllers/mattermost/clusterinstallation/mattermost.go
@@ -92,7 +92,7 @@ func (r *ClusterInstallationReconciler) checkMattermostRBAC(mattermost *mattermo
 	}
 	err = r.checkMattermostRoleBinding(mattermost, roleName, saName, reqLogger)
 	if err != nil {
-		return errors.Wrap(err, "failed to check mattermost Role")
+		return errors.Wrap(err, "failed to check mattermost RoleBinding")
 	}
 
 	return nil
@@ -208,7 +208,7 @@ func (r *ClusterInstallationReconciler) checkMattermostDeployment(mattermost *ma
 
 	err = r.checkMattermostDBSetupJob(mattermost, desired, reqLogger)
 	if err != nil {
-		return errors.Wrap(err, "failed to check mattermost DB setup job.")
+		return errors.Wrap(err, "failed to check mattermost DB setup job")
 	}
 
 	err = r.createDeploymentIfNotExists(mattermost, desired, reqLogger)
@@ -240,7 +240,7 @@ func (r *ClusterInstallationReconciler) checkMattermostDBSetupJob(mattermost *ma
 			reqLogger.Info("Creating DB setup job", "name", desiredJob.Name)
 			return r.create(mattermost, desiredJob, reqLogger)
 		}
-		return errors.Wrapf(err, "failed to get current db setup job")
+		return errors.Wrap(err, "failed to get current db setup job")
 	}
 	// For now, there is no need to perform job update, so just return.
 	return nil
--- a/controllers/mattermost/clusterinstallation/mattermost.go
+++ b/controllers/mattermost/clusterinstallation/mattermost.go
@@ -232,6 +232,7 @@ func (r *ClusterInstallationReconciler) checkMattermostDeployment(mattermost *ma
 
 func (r *ClusterInstallationReconciler) checkMattermostDBSetupJob(mattermost *mattermostv1alpha1.ClusterInstallation, deployment *appsv1.Deployment, reqLogger logr.Logger) error {
 	desiredJob := prepareJobTemplate(mattermost, deployment, mattermostApp.SetupJobName)
+	desiredJob.OwnerReferences = mattermostApp.ClusterInstallationOwnerReference(mattermost)
 
 	currentJob := &batchv1.Job{}
 	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: desiredJob.Name, Namespace: desiredJob.Namespace}, currentJob)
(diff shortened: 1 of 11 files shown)


controllers/mattermost/clusterinstallation/mattermost.go
```diff
@@ -74,6 +81,71 @@ func (r *ClusterInstallationReconciler) checkMattermostService(mattermost *matte
...
+	if err != nil {
+		return errors.Wrap(err, "failed to check mattermost ServiceAccount")
+	}
+	err = r.checkMattermostRole(mattermost, roleName, reqLogger)
+	if err != nil {
+		return errors.Wrap(err, "failed to check mattermost Role")
+	}
+	err = r.checkMattermostRoleBinding(mattermost, roleName, saName, reqLogger)
+	if err != nil {
+		return errors.Wrap(err, "failed to check mattermost Role")
```
gabrieljackson, line 95:
```suggestion
		return errors.Wrap(err, "failed to check mattermost RoleBinding")
```

controllers/mattermost/clusterinstallation/mattermost.go
```diff
@@ -132,7 +204,13 @@ func (r *ClusterInstallationReconciler) checkMattermostDeployment(mattermost *ma
 		}
 	}
 
-	desired := mattermostApp.GenerateDeployment(mattermost, dbInfo, resourceName, ingressName, imageName, minioURL)
+	desired := mattermostApp.GenerateDeployment(mattermost, dbInfo, resourceName, ingressName, saName, imageName, minioURL)
+
+	err = r.checkMattermostDBSetupJob(mattermost, desired, reqLogger)
+	if err != nil {
+		return errors.Wrap(err, "failed to check mattermost DB setup job.")
```
gabrieljackson, line 211:
```suggestion
		return errors.Wrap(err, "failed to check mattermost DB setup job")
```

controllers/mattermost/clusterinstallation/mattermost.go
```diff
@@ -152,6 +230,22 @@ func (r *ClusterInstallationReconciler) checkMattermostDeployment(mattermost *ma
...
+	desiredJob := prepareJobTemplate(mattermost, deployment, mattermostApp.SetupJobName)
+
+	currentJob := &batchv1.Job{}
+	err := r.Client.Get(context.TODO(), types.NamespacedName{Name: desiredJob.Name, Namespace: desiredJob.Namespace}, currentJob)
+	if err != nil {
+		if k8sErrors.IsNotFound(err) {
+			reqLogger.Info("Creating DB setup job", "name", desiredJob.Name)
+			return r.create(mattermost, desiredJob, reqLogger)
+		}
+		return errors.Wrapf(err, "failed to get current db setup job")
```
gabrieljackson, line 243:
```suggestion
		return errors.Wrap(err, "failed to get current db setup job")
```

pkg/mattermost/mattermost.go
```diff
@@ -232,6 +239,17 @@ func GenerateDeployment(mattermost *mattermostv1alpha1.ClusterInstallation, dbIn
...
 	envVarDB = append(envVarDB, masterDBEnvVar)
 
+	// Add init container to wait for DB setup job to complete
+	initContainers = append(initContainers, corev1.Container{
+		Name:            WaitForDBSetupContainerName,
+		Image:           "bitnami/kubectl:1.17",
+		ImagePullPolicy: corev1.PullIfNotPresent,
+		Command: []string{
+			"sh", "-c",
+			fmt.Sprintf("kubectl wait --for=condition=complete --timeout 5m job/%s", SetupJobName),
```
jwilander, line 249: Do you know how `kubectl wait` works under the covers? Is it polling the k8s api at a certain rate? My only concern would be if it polls every 30s then this could really slow down pod start-up

pkg/mattermost/mattermost.go
```diff
@@ -232,6 +239,17 @@ func GenerateDeployment(mattermost *mattermostv1alpha1.ClusterInstallation, dbIn
 
 	envVarDB = append(envVarDB, masterDBEnvVar)
 
+	// Add init container to wait for DB setup job to complete
+	initContainers = append(initContainers, corev1.Container{
```
gabrieljackson, line 243: This is probably fine where it is, but I would move it down this function until after the minio init containers are possibly attached. My thinking is all database and filestore setup should be completed before we attempt to initialize the mattermost server + database. This obviously won't affect non-minio installations, but better to have this init container go last in all cases.

gabrieljackson (changes requested):
This looks good. My only blocking request is a slight reordering of the init containers as I discuss below.

pkg/mattermost/mattermost.go
Szymongib: I would assume it runs watch on the informer so whenever the job completes it finishes quite quickly. I will double-check to make sure.

jwilander (approved):
Other than Gabe's one comment looks good to me 👍

jwilander:
@cpanato mind taking a quick look when you get a moment?

Szymongib:
I have noticed something that may cause some issues with single-tenant DB, as we do not create init containers to check if the database is ready in this scenario, which causes the job to fail a few times before the DB is ready (before adding the job it would keep crashing the MM pod).
I have increased the job backoff limit to make sure that the job will not end up in a failed state but we probably should consider adding DB checks for other scenarios than only MySQL Operator one.

@jwilander @gabrieljackson Let me know if you are ok moving forward with this approach and I will create follow-up tickets.

gabrieljackson (approved):
Agreed on adding the missing db check init containers. Just so you know, I am pretty sure we have these init check containers for all mysql options, and just need to add them for Postgres.

cpanato:
this is created only for the first time when a new installation is created? or for every update?

cpanato (approved):
did not see anything that calls my attention
/lgtm

Szymongib:
@cpanato The job will only get created when it does not exist, it will not be run on update.

PR #195, opened 2020-11-09
Add cloud specific sizes for 10 and 100 users

jwilander:
#### Summary
Looking at our production environment all our installations with 10 or less users average about 3m vCPU and 75Mi per pod so these new requests are still higher than they need to be. Even larger 100 user installations do not use significantly more CPU or memory. We can use these as the starting point for our cloud-specific sizes.

#### Release Note

```release-note
Add cloud specific sizes for 10 and 100 users
```

### DIFF
--- a/apis/mattermost/v1alpha1/clusterinstallation_sizes.go
+++ b/apis/mattermost/v1alpha1/clusterinstallation_sizes.go
@@ -57,6 +57,80 @@ var size100 = ClusterInstallationSize{
 	},
 }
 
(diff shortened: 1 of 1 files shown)


gabrieljackson (approved):
👍

Szymongib (approved):
LGTM
I think for the future it would be good to clean up the naming of different sizes a little bit as I can see it being a little confusing. Maybe a good moment to do it would be together with Operator beta spec.

stylianosrigas (approved):
LGTM

PR #196, opened 2020-11-13
MM-30448 - Support Init containers waiting for Postgres readiness

Szymongib:
#### Summary

Add the possibility to use init containers that wait for database readiness when using Postgres. Currently, it is only possible with MySQL, which makes pods crash when using Postgres and the DB is not ready.
- Init container type is decided based on the connection string.
- If `DB_CONNECTION_CHECK_URL` is not present in the Secret, the Init container is not added to the pods.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-30448

See also: https://github.com/mattermost/mattermost-cloud/pull/364

#### Release Note

```release-note
Support Init containers waiting for Postgres readiness
```

### DIFF
--- a/pkg/mattermost/mattermost.go
+++ b/pkg/mattermost/mattermost.go
@@ -159,28 +159,10 @@ func GenerateDeployment(mattermost *mattermostv1alpha1.ClusterInstallation, dbIn
 		}
 
 		if dbInfo.HasDatabaseCheckURL() {
-			initContainers = append(initContainers, corev1.Container{
-				Name:            "init-check-database",
-				Image:           "appropriate/curl:latest",
-				ImagePullPolicy: corev1.PullIfNotPresent,
-				Env: []corev1.EnvVar{
-					{
-						Name: "DB_CONNECTION_CHECK_URL",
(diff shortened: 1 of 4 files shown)


pkg/mattermost/mattermost.go
```diff
@@ -641,3 +623,48 @@ func newService(mattermost *mattermostv1alpha1.ClusterInstallation, serviceName,
...
+			ImagePullPolicy: corev1.PullIfNotPresent,
+			Env:             envVars,
+			Command: []string{
+				"sh", "-c",
+				"until pg_isready --dbname=\"$DB_CONNECTION_CHECK_URL\"; do echo waiting for database; sleep 5; done;",
+			},
+		}
+	default:
+		return nil
+	}
```
gabrieljackson, line 669: nit: could just replace this with `return nil` for the function itself.

gabrieljackson (approved):
Clean implementation. Great job!

jwilander:
>which makes pods **crush** when using Postgres and the DB is not ready

I know that's a typo but it makes it sound like the pods perform way better haha

jwilander (approved):
LGTM 👍

PR #197, opened 2020-11-18
Add v1beta1 of the Mattermost operator spec

gabrieljackson:
This new spec sets the foundation for future improvements for
deploying and managing Mattermost. Lessons learned from the v1alpha1
spec have been incorporated and new functionality will be supported
in future changes.

No controller logic has been added and further changes will be needed
to auto-generate and check the beta spec files.

Fixes https://mattermost.atlassian.net/browse/MM-30702

```release-note
Add v1beta1 of the Mattermost operator spec
```


apis/mattermost/v1beta1/mattermost_types.go
```diff
@@ -0,0 +1,263 @@
...
+	Version string `json:"version,omitempty"`
+	// Replicas defines the number of replicas to use for the Mattermost app
+	// servers.
+	Replicas int32 `json:"replicas,omitempty"`
+	// Optional environment variables to set in the Mattermost application pods.
+	// +optional
+	MattermostEnv []v1.EnvVar `json:"mattermostEnv,omitempty"`
+	// MattermostLicenseSecret is the name of the secret containing a license.
+	// +optional
+	MattermostLicenseSecret string `json:"mattermostLicenseSecret,omitempty"`
```
Szymongib, line 51: What do you think about skipping the `Mattermost` prefix for fields in this struct, as now we have `MattermostSpec. MattermostLicenseSecret` which seems unnecessary.

apis/mattermost/v1beta1/mattermost_types.go
```diff
@@ -0,0 +1,263 @@
...
+	// +optional
+	Size string `json:"size,omitempty"`
+
+	// Image defines the Mattermost Docker image.
+	Image string `json:"image,omitempty"`
+	// Version defines the Mattermost Docker image version.
+	Version string `json:"version,omitempty"`
+	// Replicas defines the number of replicas to use for the Mattermost app
+	// servers.
+	Replicas int32 `json:"replicas,omitempty"`
```
Szymongib, line 45: I would consider making it `*int32` as then if we set replicas to 0 it can really be 0 instead of using an artificial value like -1.
To expand on that I think it is good practice to use a pointer for optional values that do not have clear cut zero value like `int`, `bool` etc. Then when setting defaults we can clearly differentiate whether the value is not set or set to the default.

apis/mattermost/v1beta1/mattermost_types.go
```diff
@@ -0,0 +1,263 @@
...
+	// +optional
+	UseServiceLoadBalancer bool `json:"useServiceLoadBalancer,omitempty"`
+	// +optional
+	ServiceAnnotations map[string]string `json:"serviceAnnotations,omitempty"`
+	// +optional
+	ResourceLabels map[string]string `json:"resourceLabels,omitempty"`
+	// Volumes allows for mounting volumes from various sources into the
+	// Mattermost application containers.
+	// +optional
+	Volumes []*v1.Volume `json:"volumes,omitempty"`
```
Szymongib, line 67: We would also need `VolumeMounts` to specify paths to where the volumes should be mounted.

apis/mattermost/v1beta1/mattermost_types.go
```diff
@@ -0,0 +1,263 @@
...
+	// Defines the probe to check if the application is up and running.
+	// +optional
+	LivenessProbe v1.Probe `json:"livenessProbe,omitempty"`
+	// Defines the probe to check if the application is ready to accept traffic.
+	// +optional
+	ReadinessProbe v1.Probe `json:"readinessProbe,omitempty"`
+}
+
+// Database defines the database configuration for Mattermost.
+type Database struct {
```
Szymongib, line 101: I wonder what do you think about splitting the fields to two sub structs `SelfManaged/External` and `OperatorManaged` (probably would need better names). I think this would clean up the struct a little bit but I am not sure about all the implications.
The same applies to filestore.

apis/mattermost/v1beta1/mattermost_types.go
```diff
@@ -0,0 +1,263 @@
...
+
+	// External Services
+	Database      Database      `json:"database,omitempty"`
+	Filestore     Filestore     `json:"filestore,omitempty"`
+	ElasticSearch ElasticSearch `json:"elasticSearch,omitempty"`
+}
+
+// Advanced defines the advanced configuration for Mattermost. These settings
+// should be correct for most types of deployments.
+type Advanced struct {
```
Szymongib, line 80: I am thinking if we could come up with a better name as `Advanced` is not very descriptive.

PROJECT
```diff
@@ -10,6 +10,9 @@ resources:
 - group: mattermost.com
   kind: ClusterInstallation
   version: v1alpha1
+- group: mattermost
```
Szymongib, line 13: I wonder if we could use some different group name as the full resource name will be: `mattermost.mattermost.mattermost.com` 😄 
Maybe something like `cloud` would make sense? I do not really have any good idea here.

Szymongib (commented):
Looks good! I threw in a couple of small ideas.

I wonder tho if, after the migration, we should go with version 2.x as we remove the `BlueGreen` part which technically is breaking backward compatibility as we do not have a way to represent it with a new spec. Curious to hear your thoughts on that.

apis/mattermost/v1beta1/mattermost_types.go
gabrieljackson: Yeah, we discussed this a bit in the design doc. The comment was supposed to explain the intention of the object, but open to new ideas if we have them.

apis/mattermost/v1beta1/mattermost_types.go
gabrieljackson: Yes, I had the same idea and forgot to mention it in the original PR comment. I do think there is a good opportunity to do that work with the new spec.

I think it may be slightly easier to make these changes as we design the controller logic as I have a few ideas on how it could work, but feel like it may become clearer as we build out the actual deployment logic. Due to this new resource being clearly marked as not-production-ready, we are safe to iterate as we build. What do you think?

apis/mattermost/v1beta1/mattermost_types.go
gabrieljackson: Good catch. Will add that.

apis/mattermost/v1beta1/mattermost_types.go
gabrieljackson: Yep, I like this. Will push up a change.

apis/mattermost/v1beta1/mattermost_types.go
gabrieljackson: Yeah, I kept the old name, but it is a bit clunky. Let's change it.

PROJECT
gabrieljackson: Yeah, I struggled with this as well, but also struggled to figure out what the better option would be. `cloud` sorta makes sense to me, but not sure if something else would make more sense. Open to suggestions on this as well.

apis/mattermost/v1beta1/mattermost_types.go
Szymongib: Yup, I agree, I think oy would be hard to speculate the final form right now as it may be heavily impacted by the implementation.

jwilander (approved):
Looks good 👍 I agree that some of the implementation of the controller logic might affect the spec but I think this is a solid starting point

PROJECT
jwilander: Oof that's gross :P Maybe `platform`?

Szymongib (approved):

apis/mattermost/v1beta1/mattermost_types.go
cpanato: Since we are doing the v1beta1 do you think is better we drop the support for minio operator and MySQL operator? 

less things to we care, and we know there are lots of breaking changes and it is hard to maintain things that we cannot control

config/samples/mattermost_v1beta1_mattermost.yaml
```diff
@@ -0,0 +1,7 @@
+apiVersion: mattermost.mattermost.com/v1beta1
+kind: Mattermost
+metadata:
+  name: mattermost-sample
+spec:
+  # Add fields here
+  foo: bar
```
cpanato, line 7: this is a valid field?

cpanato (commented):
we dont need to write the conversion from alpha to beta?

jwilander:
>we dont need to write the conversion from alpha to beta?

@cpanato that's coming later - this is just to get the spec ready so we can start implementing the overall changes

config/samples/mattermost_v1beta1_mattermost.yaml
gabrieljackson: That's just a sample config that was auto-generated. As I mentioned in the original PR description, more work is needed to finalize the auto-generated files that will come later.

cpanato: sgtm, but we can remove this to not have these invalid fields in the repo, but it is up to you

cpanato (approved):
for beta I would like to drop the mysql/minio operator support

gabrieljackson:
> for beta I would like to drop the mysql/minio operator support

I wouldn't be opposed to dropping that support as well. We can discuss this as we continue to develop.

PR #200, opened 2020-12-02
MM-30827 - Adjust beta types, generate new clients and CRD

Szymongib:
#### Summary
First PR in my attempt to split the implementation of Operator Beta spec into smaller pieces.
Changes:
- Adjust types definitions
- Provide utility methods for types
- Generate client for new Mattermost type
- Generate CRD
- Adjust Makefile and CI configuration

#### Ticket Link

First part of https://mattermost.atlassian.net/browse/MM-30827

#### Release Note

```release-note
NONE
```


cpanato (approved):

apis/mattermost/v1beta1/db_util.go
```diff
@@ -0,0 +1,83 @@
...
+
+// SetDefaults sets the missing values in Database to the default ones.
+func (db *Database) SetDefaults() {
+	if db.IsExternal() {
+		return
+	}
+
+	if db.OperatorManaged == nil {
+		db.OperatorManaged = &OperatorManagedDatabase{}
+	}
```
gabrieljackson, line 18: Looks like this could be replaced with `db.ensureDefault()`?

apis/mattermost/v1beta1/mattermost_types.go
```diff
@@ -65,23 +66,31 @@ type MattermostSpec struct {
...
+	// for below settings, unless a very specific use case arises.
+
+	// Scheduling defines the configuration related to scheduling of the Mattermost pods
+	// as well as resource constraints. These settings generally don't need to be changed.
+	// +optional
+	Scheduling Scheduling `json:"scheduling,omitempty"`
+	// Probes defines configuration of liveness and readiness probe for Mattermost pods.
+	// These settings generally don't need to be changed.
+	// +optional
+	Probes Probes `json:"probes,omitempty"`
```
gabrieljackson, line 88: Hmm, I guess it's fine to split these out and use the comment to define them as "advanced" and likely to change, but the original design did put them in their own struct to make this a bit clearer. I am good with this, but we may want to revisit this again in the future before the beta spec is production-ready.

gabrieljackson (approved):
Looks great! I really like the cleanup of the filestore and database structs.

apis/mattermost/v1beta1/mattermost_types.go
Szymongib: I am 1/5 on that. I just did not like that the struct was called `Advanced` as it seems very vague but I couldn't find a well-fitting name that would work for all fields inside, therefore I decided to split it.
I am definitely open to discussing it further.

jwilander (approved):

PR #206, opened 2021-01-05
MM-31335 - Implement migration from ClusterInstallation to Mattermost

Szymongib:
#### Summary

This PR introduces migration mechanism from `ClusterInstallation` resources to `Mattermost` resources.
To migrate the resource `spec.migrate` needs to be set to `true`.
Mechanism overview:
- If `spec.migrate` is set to `true` the resource will not follow the regular reconciliation process - updating resources etc.
- If `ClusterInstallation` uses Blue Green or Canary the migration will not be performed and status containing such information will be set.
- If an error occurred during the migration, for most cases setting `spec.migrate` to false should restore state preceding the migration.
- Migration does not cause downtime to the Mattermost instance. 
- After the migration is done the `ClusterInstallation` is deleted.

Detailed documentation will be provided in other PRs.

#### Release Note

```release-note
Add mechanism to migrate from ClusterInstallation to Mattermost resource.
```


controllers/mattermost/clusterinstallation/migration.go
```diff
@@ -0,0 +1,264 @@
...
+			RequeueIn: 10 * time.Second,
+			Status:    "Migration to Mattermost is in progress - waiting for Mattermost to be ready",
+		}, nil
+	}
+
+	logger.Info("Migration finished. Removing old Replica Sets and ClusterInstallation")
+	err = r.cleanupReplicaSets(ci)
+	if err != nil {
+		return MigrationResult{}, errors.Wrap(err, "failed to cleanup old Replica Sets")
+	}
```
gabrieljackson, line 72: I am curious if this is needed due to this being a migration and there are differences from the normal ClusterInstallation resource deletion which happens immediately after this? This is normally handled for us when that deletion occurs, but I am assuming there is a difference here with the migration logic underway.

gabrieljackson (approved):
Very nice work here. No major issues or concerns.

controllers/mattermost/clusterinstallation/migration.go
Szymongib: Yes, that is due to difference in how we are handling the migration. 
To avoid downtime we need to delete old Deployment orphaning its Replica Sets so that Pods keep running until we create new Deployment (this is needed as we change the immutable `selector` field). 
In usual scenario the Replica Sets have Owner References to the Deployment and Deployment to the ClusterInstallation therefor removing ClusterInstallation cleans out the whole thing. However if we remove deployment earlier and orphan Replica Sets, there is no longer Owner Reference that triggers their removal so we need to clean the up manually.

controllers/mattermost/clusterinstallation/conversion.go
```diff
@@ -0,0 +1,169 @@
...
+	"github.com/mattermost/mattermost-operator/pkg/utils"
+	"github.com/pkg/errors"
+	corev1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+)
+
+func (r *ClusterInstallationReconciler) IsConvertible(ci *mattermostv1alpha1.ClusterInstallation) error {
+	if ci.Spec.BlueGreen.Enable {
+		return errors.New("cluster installation with BlueGreen enabled cannot be converted to Mattermost")
```
jwilander, line 18: This might make it a bit clearer if someone runs into this in the logs

```suggestion
		return errors.New("ClusterInstallation resource with BlueGreen enabled cannot be converted to Mattermost resource. Disable BlueGreen to enable migration.")
```

jwilander (approved):
Looks good 👍 just one suggestion on making a log message clearer

stylianosrigas (approved):

PR #208, opened 2021-01-13
MM-31880 - Decrease the requests of small cloud cluster installations

stafot:
Decrease the requests of small cloud cluster installations.

#### Summary
Decrease the requests of small cloud cluster installations.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-31880
#### Release Note

```release-note
NONE
```

### DIFF
--- a/apis/mattermost/v1alpha1/clusterinstallation_sizes.go
+++ b/apis/mattermost/v1alpha1/clusterinstallation_sizes.go
@@ -2,7 +2,6 @@ package v1alpha1
 
 import (
 	"github.com/pkg/errors"
-
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/resource"
 )
@@ -32,7 +31,7 @@ var size100 = ClusterInstallationSize{
 				corev1.ResourceMemory: resource.MustParse("256Mi"),
 			},
 			Limits: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("2"),
+				corev1.ResourceCPU:    resource.MustParse("2000m"),
 				corev1.ResourceMemory: resource.MustParse("4Gi"),
 			},
 		},
@@ -66,10 +65,10 @@ var cloudSize10 = ClusterInstallationSize{
 		Resources: corev1.ResourceRequirements{
 			Requests: corev1.ResourceList{
 				corev1.ResourceCPU:    resource.MustParse("10m"),
-				corev1.ResourceMemory: resource.MustParse("200Mi"),
+				corev1.ResourceMemory: resource.MustParse("150Mi"),
 			},
 			Limits: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("2"),
+				corev1.ResourceCPU:    resource.MustParse("2000m"),
 				corev1.ResourceMemory: resource.MustParse("4Gi"),
 			},
 		},
@@ -106,7 +105,7 @@ var cloudSize100 = ClusterInstallationSize{
 				corev1.ResourceMemory: resource.MustParse("256Mi"),
 			},
 			Limits: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("2"),
+				corev1.ResourceCPU:    resource.MustParse("2000m"),
 				corev1.ResourceMemory: resource.MustParse("4Gi"),
 			},
 		},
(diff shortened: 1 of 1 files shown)


apis/mattermost/v1alpha1/clusterinstallation_sizes.go
```diff
@@ -60,16 +61,20 @@ var size100 = ClusterInstallationSize{
 // CloudSize10String represents estimated Mattermost Cloud installation sizing for 10 users.
 const CloudSize10String = "cloud10users"
 
+
+var cpuInstallationRequests10u = getEnv("MM10_CPU_REQUESTS", "10m")
+var memoryInstallationRequests10u = getEnv("MM10_MEM_REQUESTS", "150Mi")
```
angeloskyratzakos, line 66: I would suggest the naming to be 
```suggestion
var cpuInstallationRequests10u = getEnv("MM_CPU_REQUESTS_10", "10m")
var memoryInstallationRequests10u = getEnv("MM_MEM_REQUESTS_10", "150Mi")
```
As firstly when I saw I thought it was for licensing purpose like the E10 or E20

apis/mattermost/v1alpha1/clusterinstallation_sizes.go
```diff
@@ -96,17 +101,19 @@ var cloudSize10 = ClusterInstallationSize{
 
 // CloudSize100String represents estimated Mattermost Cloud installation sizing for 100 users.
 const CloudSize100String = "cloud100users"
+var cpuInstallationRequests100u = getEnv("MM100_CPU_REQUESTS", "50m")
+var memoryInstallationRequests100u = getEnv("MM100_MEM_REQUESTS", "256Mi")
```
angeloskyratzakos, line 105: Same here, although it doesn't remind me the licensing, just to follow the pattern

apis/mattermost/v1alpha1/clusterinstallation_sizes.go
stafot: @angeloskyratzakos Make sense thanks for the comment.

angeloskyratzakos: Also why to have the number within the variable as in the future if we change the `MM_CPU_REQUESTS_10` to be `20m` it wouldn't make sense

stafot: @angeloskyratzakos Is a default value when ENV is not set check the helper function added to the bottom

angeloskyratzakos: maybe something like `MM_SIZE10_CPU_REQUESTS`

stafot: that's even better

apis/mattermost/v1alpha1/clusterinstallation_sizes.go
```diff
@@ -97,16 +101,19 @@ var cloudSize10 = ClusterInstallationSize{
...
+
 var cloudSize100 = ClusterInstallationSize{
 	App: ComponentSize{
 		Replicas: 2,
 		Resources: corev1.ResourceRequirements{
 			Requests: corev1.ResourceList{
-				corev1.ResourceCPU:    resource.MustParse("50m"),
-				corev1.ResourceMemory: resource.MustParse("256Mi"),
+				corev1.ResourceCPU:    resource.MustParse(cpuInstallationRequests100u),
+				corev1.ResourceMemory: resource.MustParse(cpuInstallationRequests100u),
```
gabrieljackson, line 113:
```suggestion
				corev1.ResourceMemory: resource.MustParse(memoryInstallationRequests100u),
```

gabrieljackson (changes requested):
I am slightly hesitant to go with this approach of overwriting size values with env vars.

The size value is considered a starting point and then you can tweak any resource value to your liking afterwards by setting it directly in the custom resource. Is there a reason this behavior can't provide us with what we need?

jwilander:
I'd agree with @gabrieljackson on that. Any reason we can't just tweak the resource limits/requests directly in the manifest for the installation?

stafot:
The main purpose of this PR is to reduce the baseline requests based on monitored usage since will give more room to improve efficiency. Since you both @gabrieljackson and @jwilander believe that we don't need environment variables I have no problem to remove them. My understanding was that this way (with these env vars) we can set a bigger value on a specific installation permanently if it needs it, but I might missing something on the flow and my understanding is not correct.

gabrieljackson (approved):
LGTM

jwilander (approved):

PR #212, opened 2021-01-20
MM-29613 - Set MM_FILESETTINGS_AMAZONS3SSL to true when using external file store

Szymongib:
#### Summary

`MM_FILESETTINGS_AMAZONS3SSL` will be set to `true` by default when using the external file store and to `false` in case Minio managed by the Minio Operator. User can still override it by explicitly setting `MM_FILESETTINGS_AMAZONS3SSL` in the Mattermost `spec.MattermostEnv` array.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-29613

#### Release Note

```release-note
Breaking change: Set MM_FILESETTINGS_AMAZONS3SSL to true when using external file store
```

### DIFF
--- a/pkg/mattermost/env_var.go
+++ b/pkg/mattermost/env_var.go
@@ -1,6 +1,10 @@
 package mattermost
 
-import corev1 "k8s.io/api/core/v1"
+import (
+	"strconv"
+
+	corev1 "k8s.io/api/core/v1"
+)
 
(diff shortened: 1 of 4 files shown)


jwilander (approved):
Looks good 👍 Let's make sure to call it out as a release note since it's technically a breaking change

gabrieljackson (approved):
Good tests and +1 on calling this out in the release notes.

PR #238, opened 2021-08-15
Set Observed Generation status value

gabrieljackson:
This change adds an observed generation value to the status object
of the v1beta1 mattermost resource. This value can be used to
determine if the controller has taken action on the latest
generation of the resource or not when the resource state is stable.

This can be used by external clients to ensure that the desired
changes have been properly applied which will avoid possible
race conditions.

Fixes https://mattermost.atlassian.net/browse/MM-37903

```release-note
Set Observed Generation status value
```


Szymongib (commented):
Great improvement! I think it is very useful, but I believe we need to be a little more precise with handling the new field.

Szymongib (approved):

PR #249, opened 2021-10-04
Introduce new Ingress configuration

### LINKED ISSUE #235: Allow disabling of Ingress creation
#### Summary
Currently the operator/CRD do not allow creation of a Mattermost instance without an ingress. The value `spec.ingressName` is always required to the best of my knowledge.

Our use case for this is deploying MM without an ingress and using "alternative ingress" via a service mesh and would prefer that this ingress is not created.

#### Possible fixes
Provide a new value (defaulted to true) that toggles the creation of Ingress. When false, it would also "disable the requirement" to provide a `spec.ingressName`. There could be numerous ways to implement this, but probably the most backwards compatible would be with a new `spec.ingressEnabled` flag:

```yaml
apiVersion: installation.mattermost.com/v1beta1
kind: Mattermost
...
spec:
  ingressEnabled: false # defaults to true
...
```

If open to changing the spec slightly, it would seem to make more logical sense and better match other operators to do it with a `spec.ingress` block:
```yaml
apiVersion: installation.mattermost.com/v1beta1
kind: Mattermost
...
spec:
  ingress:
    enabled: true
    name: my-mm-ingress
    annotations:
      kubernetes.io/ingress.class: nginx
...
```

Totally understand if the second is off the table because it would introduce "breaking" changes. Either of these proposed solutions (or really anything that allows us to disable the Ingress) would work for our needs.

Szymongib:
#### Summary

This PR introduces a new Ingress configuration to Mattermost Custom Resource.
The ingress configuration is now moved to the `ingress` subsection of the spec:
- `ingressName` -> `ingress.host`
- `ingressAnnotations` -> `ingress.annotations`
- `useIngressTLS` -> `ingress.tlsSecret`

Additionally: 
- Ingress creation can now be disabled by setting `spec.ingress.enabled` to `false`.
- TLS secret for the Ingress can be specified with `spec.ingress.tlsSecret` instead of using a hardcoded value.

The old fields will continue to work until the next CR version. The new configuration (if specified) will take precedence over old fields.

Example CR:
```yaml
apiVersion: installation.mattermost.com/v1beta1
kind: Mattermost
metadata:
  name: mm-test
  namespace: mm-test
spec:
  image: mattermost/mattermost-enterprise-edition
  imagePullPolicy: IfNotPresent
  ingress:
    annotations:
      test: test
    enabled: true
    host: test.com
    tlsSecret: test
  version: 5.37.1
```

#### Ticket Link
https://mattermost.atlassian.net/browse/MM-38985

Fixes #235

#### Release Note

```release-note
Introduce new Ingress configuration
Allow disabling creation of Ingress
Allow specifying secret for TLS cert
```


emilyacook:
Thanks for participating in Hacktoberfest! You can claim your sticker set here: https://get.printfection.com/hacktober21/4144583267 @Szymongib

PR #262, opened 2021-11-04
Do not set SiteURL if ingress host is not provided

### LINKED ISSUE #261: Disabling ingress still requires `ingress.host`
#### Summary

The new `ingress.enabled` spec does allow users to disable ingress but it seems like the `ingress.host` value in the spec is used for much more than just the ingress. Disabling ingress without providing an "accurate" value to `ingress.host` causes a number of issues.

#### Steps to reproduce
1. Deploy the operator.
2. Deploy a Mattermost resource with `ingress.enabled` set to false and nothing set under `ingress.host` or `ingressName`.
3. Attempt to navigate to the UI via alternative means (i.e. port-forward or using a service mesh ingress).
4. You should see the UI load a single white screen, with some errors in the browser console.

#### Expected behavior
Ingress should be able to be fully disabled with the UI still working.

Ingress is able to be disabled, but unless you provide an `ingress.host` the UI does not function.

#### Possible fixes

I'm not 100% sure but I suspect it is due to a line [here](https://github.com/mattermost/mattermost-operator/blob/master/pkg/mattermost/mattermost.go#L310) where it is used for the siteUrl - which [affects things in the `config.json`](https://github.com/mattermost/mattermost-operator/blob/master/pkg/mattermost/env_var.go#L9).

Assuming this is the spot that causes the issue, this is clearly a needed variable/config to set. Perhaps the solution is to do a number of things:
1. A new `spec.siteURL` (name/place wherever it makes sense) that is used to populate the `config.json` when available, but defaults to `ingress.host` when not set.
2. Error handling for the operator to reflect this flow ^. Since the ingress was made optional the operator should perform some validation that the siteURL is populated with a value otherwise users can deploy, pods will go to running, but the UI will be unusable.
3. Better document this (maybe not needed if you add a new `siteUrl` value). Currently examples mention that [ingressName is required](https://github.com/mattermost/mattermost-operator/blob/master/docs/examples/mattermost_full.yaml#L17), but only when using an ingress.

While there is a workaround, I think the naming of the values in spec could get confusing and at minimum this should be documented.

Szymongib:
#### Summary

The issue manifests when Ingress is disabled or a Service LoadBalancer is used.

We should not set `MM_SERVICESETTINGS_SITEURL` env var when ingress host is not provided as this prevents MM from working properly.

Instead, the `SiteURL` will not be set which results in prompting the administrator to set it in System Console.
The `SiteURL` can still be set in the manifest by explicitly providing env var.

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-39871

Fixes #261

#### Release Note

```release-note
Do not set SiteURL if ingress host is not provided
```

### DIFF
--- a/pkg/mattermost/env_var.go
+++ b/pkg/mattermost/env_var.go
@@ -7,11 +7,7 @@ import (
 )
 
 func generalMattermostEnvVars(siteURL string) []corev1.EnvVar {
-	return []corev1.EnvVar{
-		{
-			Name:  "MM_SERVICESETTINGS_SITEURL",
-			Value: siteURL,
-		},
+	envs := []corev1.EnvVar{
 		{
 			Name:  "MM_PLUGINSETTINGS_ENABLEUPLOADS",
 			Value: "true",
@@ -37,6 +33,15 @@ func generalMattermostEnvVars(siteURL string) []corev1.EnvVar {
 			Value: "kubernetes-operator",
 		},
 	}
+
+	if siteURL != "" {
+		envs = append(envs, corev1.EnvVar{
+			Name:  "MM_SERVICESETTINGS_SITEURL",
+			Value: siteURL,
+		})
+	}
+
+	return envs
 }
 
 func fileStoreEnvVars(fileStore *FileStoreInfo) []corev1.EnvVar {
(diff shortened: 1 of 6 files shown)


gabrieljackson (approved):
👍

PR #264, opened 2021-12-10
Add possibility to specify resource patch

Szymongib:
#### Summary

This PR introduces a new mechanism of customizing Kubernetes resources created by Operator with JSON patch.

We want to avoid introducing every possible customization to `Mattermost` CR yet there are a lot of different edge cases that may require it. This mechanism can eliminate a need for putting everything in the CR manifest while giving users a high degree of flexibility.

The downside of using this feature is that an incorrect patch can possibly break the functionality of Mattermost, or some of the K8s resources, or may prevent resources from being updated. 

Example patch opening a port on Deployment and Service:
```yaml
apiVersion: installation.mattermost.com/v1beta1
kind: Mattermost
metadata:
  name: mm-example-anywhere
spec:
...
resourcePatch:
  deployment:
    disable: false
    patch: |
      [
          {
              "op":"add",
              "path":"/spec/template/spec/containers/0/ports/-",
              "value":{"containerPort":8443, "name":"calls", "protocol":"UDP"}
          }
      ]
  service:
    disable: false
    patch: '[{"op":"add","path":"/spec/ports/-", "value": {"name": "calls", "port": 8443}}]'
```

#### Ticket Link

Open-source Friday

#### Release Note

```release-note
Add possibility to specify resource patch
```


gabrieljackson (approved):
Awesome!

PR #267, opened 2021-12-14
Add `appProtocol` to Service Creation

### LINKED ISSUE #266: Service port standard naming for Istio
#### Summary

Currently the service that is created by the operator has two ports, `app` and `metrics` (code [ref](https://github.com/mattermost/mattermost-operator/blob/master/pkg/mattermost/mattermost_v1beta.go#L73)).

Istio (service mesh) utilizes the port names for [automatic protocol detection](https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/) for its sidecars. The important part here is `If the protocol cannot automatically be determined, traffic will be treated as plain TCP traffic`. I ran across this issue with monitoring Mattermost metrics. When you point a prometheus instance at the Mattermost metrics port, it returns a 404 (if Prometheus has Istio sidecars for traffic) due to the "non-standard" port name.

#### Steps to reproduce

This will be a challenging one to reproduce without knowledge and ability to deploy Istio, Prometheus, and Mattermost. I'm happy to walk through a deployment where I'm seeing this issue if that'd be beneficial (video call or similar might be best for this given the amount of detail and complexity).

#### Expected behavior

I would expect that Mattermost metrics can be monitored in any situation, including when my Prometheus instance is Istio injected.

Prometheus returns 404s on the Mattermost endpoints.

#### Possible fixes

I think there's a couple routes that could be taken. For initial testing to validate that this is the issue I worked on [this branch](https://github.com/mjnagel/mattermost-operator/tree/service-port-name), with the changes essentially just renaming the two port names in the service generation (to `http-app` and `http-metrics` so that Istio knows they are HTTP). Confirmed that this works, the only changes required are to the "external" port names, the `targetPort` doesn't need to be changed.

I'd assume this wouldn't be an ideal solution since it could break things for someone potentially. I think ideally the operator provides a way to override the default names of these ports. Example of a simple way to do this:
```yaml
spec:
  service:
    appPortName: "http-app" # defaults to app
    metricsPortName: "http-metrics" # defaults to metrics
```

If we wanted to go further than that and allow more customization on the service (if I'm remember right `serviceAnnotations` are already a value?), we could refactor things into a `service` section in the spec similar to what was done with `Ingress` such as:
```yaml
spec:
  service:
    annotations: ...
    ports:
      - name: "http-app"
         port: 8065
      - name: "http-metrics"
         port: 8067
```

I don't know if that exactly makes sense to do but hopefully that gives an idea of what I'm thinking. Ultimately just looking for a way that makes sense to allow port name overriding and is maintainable/extensible for the future.

mjnagel:
#### Summary

Adds `appProtocol` to the service definition which helps especially with Istio injection so that sidecars know the right protocol.

#### Ticket Link

Fixes #266 

#### Release Note

```release-note
Service is now created with appProtocol defined
```

### DIFF
--- a/pkg/mattermost/mattermost_v1beta.go
+++ b/pkg/mattermost/mattermost_v1beta.go
@@ -54,14 +54,16 @@ func GenerateServiceV1Beta(mattermost *mmv1beta.Mattermost) *corev1.Service {
 func configureMattermostLoadBalancerService(service *corev1.Service) *corev1.Service {
 	service.Spec.Ports = []corev1.ServicePort{
 		{
-			Name:       "http",
-			Port:       80,
-			TargetPort: intstr.FromString("app"),
+			Name:        "http",
+			Port:        80,
+			AppProtocol: pkgUtils.NewString("http"),
+			TargetPort:  intstr.FromString("app"),
 		},
 		{
-			Name:       "https",
-			Port:       443,
-			TargetPort: intstr.FromString("app"),
+			Name:        "https",
+			Port:        443,
+			AppProtocol: pkgUtils.NewString("https"),
+			TargetPort:  intstr.FromString("app"),
 		},
 	}
 	service.Spec.Type = corev1.ServiceTypeLoadBalancer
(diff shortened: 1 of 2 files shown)


Szymongib (commented):
Thanks a lot for adding this!

Would you mind adding a simple check to tests to avoid regressions?

I think we could add `appProtocol` param to this function and check it together with port number https://github.com/mattermost/mattermost-operator/blob/master/pkg/mattermost/mattermost_v1beta_test.go#L41-L49.

mjnagel:
> I think we could add `appProtocol` param to this function and check it together with port number https://github.com/mattermost/mattermost-operator/blob/master/pkg/mattermost/mattermost_v1beta_test.go#L41-L49.

@Szymongib I think I got that added - let me know if that looks like what you were expecting. Split off the if conditions so that you can easily have a different Fail message depending on the issue.

Szymongib (approved):
Thanks @mjnagel! Looks great!

gabrieljackson (approved):
Awesome! Thanks for the contribution!

PR #273, opened 2022-02-04
Add e2e tests for external DB and file-store configuration

Szymongib:
#### Summary

This PR adds a new e2e test that uses external database and file store configuration instead of Minio and MySQL operators. 
Test deploys Postgres and MinIO locally in the cluster but those are used as an external in CR configuration. 

Tests are more lightweight as they do not require other operators. PR provides a simple framework to easily set up additional tests in isolated namespaces. We can leverage that to implement more e2e tests for more specific configuration options. 

#### Release Note

```release-note
Add e2e tests for external DB and file-store configuration
```


gabrieljackson (approved):
Awesome!

PR #283, opened 2022-05-27
fix: Add owner reference update-check job

### LINKED ISSUE #215: Add owner reference to update-check job so that it points to Mattermost resource
## Description

When modifying `spec.image` or `spec.version` of Mattermost resource the update-check job is run to ensure image is correct.
However, the job is not owned by the Mattermost resource which causes Operator to not to pick up when the job completes and carry on with update.

### Possible solution

Very simple and reliable solution would be to add `OwnerReference` to the job, so that it is owned by Mattermost resource.
We already have [function](https://github.com/mattermost/mattermost-operator/blob/272928435a67a730fe50583fee309a738c9d8536/pkg/resources/create_resources.go#L48) that adds owner references when creating objects. We could use it [when lunching the update job](https://github.com/mattermost/mattermost-operator/blob/272928435a67a730fe50583fee309a738c9d8536/pkg/resources/update_job.go#L19)

imskr:
#### Summary

Add `OwnerReference` when updating job

#### Ticket Link

Fixes #215 

#### Release Note

```release-note
Add owner reference update-check job
```

### DIFF
--- a/controllers/mattermost/mattermost/mattermost_test.go
+++ b/controllers/mattermost/mattermost/mattermost_test.go
@@ -278,6 +278,8 @@ func TestCheckMattermost(t *testing.T) {
 		// Assert owner references are set here
 		assert.NotEmpty(t, job.ObjectMeta.OwnerReferences)
 		assert.Equal(t, "Mattermost", job.ObjectMeta.OwnerReferences[0].Kind)
+		assert.Equal(t, mmName, job.ObjectMeta.OwnerReferences[0].Name)
+		assert.Equal(t, "installation.mattermost.com/v1beta1", job.ObjectMeta.OwnerReferences[0].APIVersion)
 
 		// Set job status to succeeded so that test can proceed
 		now := metav1.Now()
--- a/controllers/mattermost/mattermost/mattermost_test.go
+++ b/controllers/mattermost/mattermost/mattermost_test.go
@@ -3,12 +3,11 @@ package mattermost
 import (
 	"context"
 	"fmt"
-	"testing"
-
 	"github.com/go-logr/logr"
 	"github.com/sirupsen/logrus"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"testing"
 
 	"github.com/mattermost/mattermost-operator/pkg/resources"
 
@@ -237,27 +236,7 @@ func TestCheckMattermost(t *testing.T) {
 	})
 
 	t.Run("deployment", func(t *testing.T) {
-		updateName := "mattermost-update-check"
-		now := metav1.Now()
-		job := &batchv1.Job{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      updateName,
-				Namespace: mm.GetNamespace(),
-			},
-			Spec: batchv1.JobSpec{Template: corev1.PodTemplateSpec{
-				Spec: corev1.PodSpec{
-					Containers: []corev1.Container{
-						{Name: mmv1beta.MattermostAppContainerName, Image: mm.GetImageName()},
-					},
-				},
-			}},
-			Status: batchv1.JobStatus{
-				Succeeded:      1,
-				CompletionTime: &now,
-			},
-		}
-		err = reconciler.Client.Create(context.TODO(), job)
-		require.NoError(t, err)
+		updateJobName := "mattermost-update-check"
 
 		err = reconciler.checkMattermostDeployment(mm, dbInfo, fileStoreInfo, currentMMStatus, logger)
 		assert.NoError(t, err)
@@ -285,8 +264,31 @@ func TestCheckMattermost(t *testing.T) {
 
 		err = reconciler.Client.Update(context.TODO(), modified)
 		require.NoError(t, err)
+
+		// Update job should be launched, we expect error
+		err = reconciler.checkMattermostDeployment(mm, dbInfo, fileStoreInfo, currentMMStatus, logger)
+		require.Error(t, err)
+
+		// Assert the job was created
+		job := &batchv1.Job{}
+		err = reconciler.Client.Get(context.TODO(), types.NamespacedName{Name: updateJobName, Namespace: mmNamespace}, job)
+		require.NoError(t, err)
+
+		// TODO: assert owner references are set here
+
+		// Set job status to succeeded so that test can proceed
+		now := metav1.Now()
+		job.Status =  batchv1.JobStatus{
+			Succeeded:      1,
+			CompletionTime: &now,
+		}
+		err = reconciler.Client.Update(context.TODO(), job)
+		require.NoError(t, err)
+
+		// Job is marked as succeeded, should proceed now.
 		err = reconciler.checkMattermostDeployment(mm, dbInfo, fileStoreInfo, currentMMStatus, logger)
 		require.NoError(t, err)
+
 		err = reconciler.Client.Get(context.TODO(), types.NamespacedName{Name: mmName, Namespace: mmNamespace}, found)
 		require.NoError(t, err)
 		assert.Equal(t, original.GetLabels(), found.GetLabels())
--- a/controllers/mattermost/mattermost/mattermost_test.go
+++ b/controllers/mattermost/mattermost/mattermost_test.go
@@ -3,11 +3,12 @@ package mattermost
 import (
 	"context"
 	"fmt"
+	"testing"
+
 	"github.com/go-logr/logr"
 	"github.com/sirupsen/logrus"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
-	"testing"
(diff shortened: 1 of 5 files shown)


gabrieljackson:
Hey @imskr thanks for the contribution! It looks like we may have a failing e2e test as a result of the change. I was only able to skim the logs, but will have more time to investigate tomorrow.

pkg/resources/update_job.go
```diff
@@ -19,10 +20,11 @@ const UpdateJobName = "mattermost-update-check"
 func (r *ResourceHelper) LaunchMattermostUpdateJob(
 	jobNamespace string,
 	baseDeployment *appsv1.Deployment,
+	reqLogger logr.Logger,
 ) error {
 	job := PrepareMattermostJobTemplate(UpdateJobName, jobNamespace, baseDeployment)
 
-	err := r.client.Create(context.TODO(), job)
+	err := r.Create(job, baseDeployment, reqLogger)
```
Szymongib, line 27: This will actually attempt to create a deployment with owner reference to the job. This is perhaps a bit unintuitive but for legacy reasons the first argument to the function is actually the object owner and the second is object to create, take a look at [the function definition](https://github.com/mattermost/mattermost-operator/blob/master/pkg/resources/create_resources.go#L47-L61).

Also we would probably like, the `Mattermost` custom resource to be the owner, so we might need to pass it to the `LaunchMattermostUpdateJob` function.

Szymongib (commented):
Thanks for the PR!

I have added suggestion that might be the cause of e2e failure, you can also try running the tests locally with `./test/e2e_local.sh` - https://github.com/mattermost/mattermost-operator/blob/master/test/e2e_local.sh

pkg/resources/update_job.go
imskr: Thanks so much @Szymongib

Szymongib: @imskr It actually looks like build issue caught early by the linter. Have you tried running unit tests locally? Try `make unittest`?

gabrieljackson: @imskr yeah it looks like you changed the `Reconcile` method signature: https://github.com/mattermost/mattermost-operator/pull/283/files#diff-0abb08611cc6662ee819cd601d4402335cce0426c2e20ecbc30c7edf1ddd78f4R64

This signature shouldn't be modified as it is a very low level part of the operator that fetches a clusterinstallation custom resource based on the provided context. https://github.com/mattermost/mattermost-operator/blob/master/controllers/mattermost/clusterinstallation/controller.go#L67

It may be a little confusing, but we have two main controllers here for different versions of a mattermost custom resource. The one seen here is the original one and deals with `mattermostv1alpha1.ClusterInstallation` and not the `v1beta1.Mattermost` resource. As such, we shouldn't really pass that in anywhere.

imskr:
Thanks for all the help @gabrieljackson @Szymongib It's fixed :)

gabrieljackson (approved):
Looks good! Thanks for the contribution.

Szymongib (commented):
Thanks for your contribution! Overall looks good, it looks however that we have a bug in the old code in `Create` method, and owner references are set after the creation, which results in the job still missing the owner reference https://github.com/mattermost/mattermost-operator/blob/master/pkg/resources/create_resources.go#L55-L60

I think we can try to correct it and see if it does not break anything else, otherwise we should add onwer references eplicitly [here](https://github.com/mattermost/mattermost-operator/blob/master/pkg/resources/update_job.go#L70-L84). We can use existing method for creating those [as we do for other resources](https://github.com/mattermost/mattermost-operator/blob/master/pkg/mattermost/mattermost_v1beta.go#L121) - this might be harder for the legacy `ClusterInstallation` resource, however.

Additionally I think it would be good to adjust tests, so that we can verify it. I was thinking we can adjust existing tests [here](https://github.com/mattermost/mattermost-operator/blob/master/controllers/mattermost/mattermost/mattermost_test.go#L239) with something like:
```
diff --git a/controllers/mattermost/mattermost/mattermost_test.go b/controllers/mattermost/mattermost/mattermost_test.go
index ce683ff5..f8781241 100644
--- a/controllers/mattermost/mattermost/mattermost_test.go
+++ b/controllers/mattermost/mattermost/mattermost_test.go
@@ -3,12 +3,11 @@ package mattermost
 import (
 	"context"
 	"fmt"
-	"testing"
-
 	"github.com/go-logr/logr"
 	"github.com/sirupsen/logrus"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	"sigs.k8s.io/controller-runtime/pkg/client"
+	"testing"
 
 	"github.com/mattermost/mattermost-operator/pkg/resources"
 
@@ -237,27 +236,7 @@ func TestCheckMattermost(t *testing.T) {
 	})
 
 	t.Run("deployment", func(t *testing.T) {
-		updateName := "mattermost-update-check"
-		now := metav1.Now()
-		job := &batchv1.Job{
-			ObjectMeta: metav1.ObjectMeta{
-				Name:      updateName,
-				Namespace: mm.GetNamespace(),
-			},
-			Spec: batchv1.JobSpec{Template: corev1.PodTemplateSpec{
-				Spec: corev1.PodSpec{
-					Containers: []corev1.Container{
-						{Name: mmv1beta.MattermostAppContainerName, Image: mm.GetImageName()},
-					},
-				},
-			}},
-			Status: batchv1.JobStatus{
-				Succeeded:      1,
-				CompletionTime: &now,
-			},
-		}
-		err = reconciler.Client.Create(context.TODO(), job)
-		require.NoError(t, err)
+		updateJobName := "mattermost-update-check"
 
 		err = reconciler.checkMattermostDeployment(mm, dbInfo, fileStoreInfo, currentMMStatus, logger)
 		assert.NoError(t, err)
@@ -285,8 +264,31 @@ func TestCheckMattermost(t *testing.T) {
 
 		err = reconciler.Client.Update(context.TODO(), modified)
 		require.NoError(t, err)
+
+		// Update job should be launched, we expect error
+		err = reconciler.checkMattermostDeployment(mm, dbInfo, fileStoreInfo, currentMMStatus, logger)
+		require.Error(t, err)
+
+		// Assert the job was created
+		job := &batchv1.Job{}
+		err = reconciler.Client.Get(context.TODO(), types.NamespacedName{Name: updateJobName, Namespace: mmNamespace}, job)
+		require.NoError(t, err)
+
+		// TODO: assert owner references are set here
+
+		// Set job status to succeeded so that test can proceed
+		now := metav1.Now()
+		job.Status =  batchv1.JobStatus{
+			Succeeded:      1,
+			CompletionTime: &now,
+		}
+		err = reconciler.Client.Update(context.TODO(), job)
+		require.NoError(t, err)
+
+		// Job is marked as succeeded, should proceed now.
 		err = reconciler.checkMattermostDeployment(mm, dbInfo, fileStoreInfo, currentMMStatus, logger)
 		require.NoError(t, err)
+
 		err = reconciler.Client.Get(context.TODO(), types.NamespacedName{Name: mmName, Namespace: mmNamespace}, found)
 		require.NoError(t, err)
 		assert.Equal(t, original.GetLabels(), found.GetLabels())
```
and we can check that owner references are set there.

One last thing would be to adjust [this method](https://github.com/mattermost/mattermost-operator/blob/master/pkg/resources/update_job.go#L34) as well, to include owner reference in recreated job.

imskr:
Thanks @Szymongib I've made the changes as suggested :)

controllers/mattermost/mattermost/mattermost_test.go
```diff
@@ -285,8 +264,31 @@ func TestCheckMattermost(t *testing.T) {
...
+		// Update job should be launched, we expect error
+		err = reconciler.checkMattermostDeployment(mm, dbInfo, fileStoreInfo, currentMMStatus, logger)
+		require.Error(t, err)
+
+		// Assert the job was created
+		job := &batchv1.Job{}
+		err = reconciler.Client.Get(context.TODO(), types.NamespacedName{Name: updateJobName, Namespace: mmNamespace}, job)
+		require.NoError(t, err)
+
+		// TODO: assert owner references are set here
```
Szymongib, line 277: I think we can address this TODO and add this check now, to make sure new functionality is working properly 🙂

Szymongib (commented):
Thanks a lot @imskr! Will have a look tomorrow.

Szymongib (commented):
Thanks a lot for making the changes!

I think we still would need to fix te bug (not related to your changes) that I mentioned above (switching the order here https://github.com/mattermost/mattermost-operator/blob/master/pkg/resources/create_resources.go#L55-L60) as without it, the owner references are not assigned.

controllers/mattermost/mattermost/mattermost_test.go
imskr: I will add this. Thanks

imskr:
Thanks @Szymongib I've added the changes :)

Szymongib (approved):
LGTM! Thanks, @imskr!

Szymongib:
@gabrieljackson since there were some changes, do you want to take another look, or you are good with merging it?

PR #285, opened 2022-05-31
Use k8s 1.22 job as default, remove the old one

Szymongib:
#### Summary

As it no longer makes sense to test with Kubernetes 1.19, we remove the old CI job and use the one with K8s 1.22 as a default.

#### Release Note

```release-note
NONE
```


gabrieljackson (approved):
Yep, this makes sense. Thanks.

PR #288, opened 2022-06-04
Don't return errors waiting for job update

### LINKED ISSUE #224: Do not return errors from reconciliation when waiting for update job
#### Summary
Before changes are rolled out to Mattermost deployment, the update job is deployed to verify that the image is correct. The Operator waits for the job to complete before starting deployment update.

From reconciler perspective, when the job is still running the error is returned to requeue reconciliation request. If error occur enough number of times, in rare cases this may significantly delay rolling out the deployment due to error reconciliation back off.

We should modify logic that waits for the update job to finish to return some indication that it is not yet finished rather than returning an error and use constant time for requeue delay. For now it should be enough to return `bool` alongside error and propagate it from [`checkUpdateJob` function](https://github.com/mattermost/mattermost-operator/blob/master/controllers/mattermost/mattermost/mattermost.go#L307) to reconciler.

imskr:
#### Summary

- Use `struct` to indicate status of update job check and sleep for 5 seconds before reconciling

#### Ticket Link

Fixes #224 

#### Release Note

```release-note
Don't return errors waiting for job update
```

### DIFF
--- a/controllers/mattermost/mattermost/controller.go
+++ b/controllers/mattermost/mattermost/controller.go
@@ -144,10 +144,18 @@ func (r *MattermostReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 		return reconcile.Result{}, err
 	}
 
-	err = r.checkMattermost(mattermost, dbConfig, fileStoreConfig, &status, reqLogger)
-	if err != nil {
-		r.updateStatusReconcilingAndLogError(mattermost, status, reqLogger, err)
-		return reconcile.Result{}, err
+	recStatus := reconcileStatus{
+		Status: false,
+		Error:  nil,
+	}
+
+	recStatus = r.checkMattermost(mattermost, dbConfig, fileStoreConfig, &status, reqLogger)
+	if recStatus.Error != nil {
+		if recStatus.Status {
+			time.Sleep(5 * time.Second)
+		}
+		r.updateStatusReconcilingAndLogError(mattermost, status, reqLogger, recStatus.Error)
+		return reconcile.Result{}, recStatus.Error
 	}
 
 	status, err = r.checkMattermostHealth(mattermost, status, reqLogger)
--- a/controllers/mattermost/mattermost/controller.go
+++ b/controllers/mattermost/mattermost/controller.go
@@ -152,7 +152,7 @@ func (r *MattermostReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 	}
 
 	if !recStatus.ResourcesReady {
-		reqLogger.Info("Mattermost resources not ready, dealying for 10 seconds!")
+		reqLogger.Info("Mattermost resources not ready, delaying for 10 seconds!")
 		return ctrl.Result{RequeueAfter: resourcesReadyDelay}, nil
 	}
 
--- a/controllers/mattermost/mattermost/controller.go
+++ b/controllers/mattermost/mattermost/controller.go
@@ -25,6 +25,7 @@ import (
 )
 
 const healthCheckRequeueDelay = 6 * time.Second
+const resourcesReadyDelay = 10 * time.Second
 
 // MattermostReconciler reconciles a Mattermost object
 type MattermostReconciler struct {
@@ -144,18 +145,15 @@ func (r *MattermostReconciler) Reconcile(ctx context.Context, request ctrl.Reque
 		return reconcile.Result{}, err
 	}
 
-	recStatus := reconcileStatus{
-		Status: false,
-		Error:  nil,
+	recStatus, err := r.checkMattermost(mattermost, dbConfig, fileStoreConfig, &status, reqLogger)
+	if err != nil {
+		r.updateStatusReconcilingAndLogError(mattermost, status, reqLogger, err)
+		return reconcile.Result{}, err
 	}
 
-	recStatus = r.checkMattermost(mattermost, dbConfig, fileStoreConfig, &status, reqLogger)
-	if recStatus.Error != nil {
-		if recStatus.Status {
-			time.Sleep(5 * time.Second)
-		}
-		r.updateStatusReconcilingAndLogError(mattermost, status, reqLogger, recStatus.Error)
-		return reconcile.Result{}, recStatus.Error
+	if !recStatus.ResourcesReady {
+		reqLogger.Info("Mattermost resources not ready, dealying for 10 seconds!")
+		return ctrl.Result{RequeueAfter: resourcesReadyDelay}, nil
 	}
 
 	status, err = r.checkMattermostHealth(mattermost, status, reqLogger)
--- a/controllers/mattermost/mattermost/mattermost.go
+++ b/controllers/mattermost/mattermost/mattermost.go
@@ -355,8 +355,11 @@ func (r *MattermostReconciler) updateMattermostDeployment(
 	}
 
 	// Job completed successfully
+	if recStatus.ResourcesReady {
+		return recStatus, r.Resources.Update(current, desired, reqLogger)
+	}
 
-	return recStatus, r.Resources.Update(current, desired, reqLogger)
+	return recStatus, nil
 }
 
 // checkUpdateJob checks whether update job status. In case job is not running it is launched
@@ -377,8 +380,8 @@ func (r *MattermostReconciler) checkUpdateJob(
 			if err = r.Resources.LaunchMattermostUpdateJob(mattermost, jobNamespace, baseDeployment, reqLogger); err != nil {
 				return nil, reconcileStatus{}, errors.Wrap(err, "Launching update image job failed")
 			}
-			reqLogger.Info("Began update image job")
 			recStatus.ResourcesReady = false
+			return nil, recStatus, errors.Wrap(err, "failed to restart update job")
 		}
 
 		return nil, reconcileStatus{}, errors.Wrap(err, "failed to determine if an update image job is already running")
@@ -398,24 +401,24 @@ func (r *MattermostReconciler) checkUpdateJob(
 		reqLogger.Info("Mattermost image changed, restarting update job")
 		err = r.Resources.RestartMattermostUpdateJob(mattermost, job, baseDeployment, reqLogger)
 		if err != nil {
-			reqLogger.Error(err, "failed to restart update job")
 			recStatus.ResourcesReady = false
+			return nil, recStatus, errors.Wrap(err, "failed to restart update job")
 		}
 
-		reqLogger.Error(err, "Restarted update image job")
 		recStatus.ResourcesReady = false
+		return nil, recStatus, errors.New("Restarted update image job")
 	}
 
 	if job.Status.CompletionTime == nil {
-		reqLogger.Error(err, "update image job still running")
 		recStatus.ResourcesReady = false
+		return nil, recStatus, errors.New("update image job still running")
 	}
 
 	// Job is completed, can check completion status
 
 	if job.Status.Failed > 0 {
-		reqLogger.Error(err, "update image job failed")
 		recStatus.ResourcesReady = false
+		return job, recStatus, errors.New("update image job failed")
 	}
 
 	reqLogger.Info("Update image job ran successfully")
--- a/controllers/mattermost/mattermost/mattermost.go
+++ b/controllers/mattermost/mattermost/mattermost.go
@@ -345,14 +345,12 @@ func (r *MattermostReconciler) updateMattermostDeployment(
 
 	reqLogger.Info("Current image is not the same as the requested, will upgrade the Mattermost installation")
 
-	job, err := r.checkUpdateJob(mattermost, mattermost.Namespace, desired, reqLogger)
+	job, recStatus, err := r.checkUpdateJob(mattermost, mattermost.Namespace, desired, reqLogger)
 	if job != nil {
-		recStatus.ResourcesReady = true
 		// Job is done, need to cleanup
 		defer r.cleanupUpdateJob(job, reqLogger)
 	}
 	if err != nil {
-		recStatus.ResourcesReady = false
 		return recStatus, err
 	}
 
@@ -367,19 +365,23 @@ func (r *MattermostReconciler) checkUpdateJob(
 	jobNamespace string,
 	baseDeployment *appsv1.Deployment,
 	reqLogger logr.Logger,
-) (*batchv1.Job, error) {
+) (*batchv1.Job, reconcileStatus, error) {
+	recStatus := reconcileStatus{
+		ResourcesReady: true,
+	}
 	reqLogger.Info(fmt.Sprintf("Running Mattermost update image job check for image %s", mmv1beta.GetMattermostAppContainerFromDeployment(baseDeployment).Image))
 	job, err := r.Resources.FetchMattermostUpdateJob(jobNamespace)
 	if err != nil {
 		if k8sErrors.IsNotFound(err) {
 			reqLogger.Info("Launching update image job")
 			if err = r.Resources.LaunchMattermostUpdateJob(mattermost, jobNamespace, baseDeployment, reqLogger); err != nil {
-				return nil, errors.Wrap(err, "Launching update image job failed")
+				return nil, reconcileStatus{}, errors.Wrap(err, "Launching update image job failed")
 			}
-			return nil, errors.New("Began update image job")
+			reqLogger.Info("Began update image job")
+			recStatus.ResourcesReady = false
 		}
 
-		return nil, errors.Wrap(err, "failed to determine if an update image job is already running")
+		return nil, reconcileStatus{}, errors.Wrap(err, "failed to determine if an update image job is already running")
 	}
 
 	// Job is either running or completed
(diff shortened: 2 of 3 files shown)


controllers/mattermost/mattermost/controller.go
```diff
@@ -144,10 +144,18 @@ func (r *MattermostReconciler) Reconcile(ctx context.Context, request ctrl.Reque
...
 	}
 
-	err = r.checkMattermost(mattermost, dbConfig, fileStoreConfig, &status, reqLogger)
-	if err != nil {
-		r.updateStatusReconcilingAndLogError(mattermost, status, reqLogger, err)
-		return reconcile.Result{}, err
+	recStatus := reconcileStatus{
+		Status: false,
+		Error:  nil,
+	}
```
Szymongib, line 150: Let's not include an error in the status, it is clearer if functions will return `(reconcileStatus, error)`.

Also, we can be a bit more specific here, perhaps instead of `Status` field, let's call it something like `ResourcesReady` or `Requeue` depending on what value we want to use as a default.

We can also remove this initialization and initialize a variable directly from the function.

controllers/mattermost/mattermost/controller.go
```diff
@@ -144,10 +144,18 @@ func (r *MattermostReconciler) Reconcile(ctx context.Context, request ctrl.Reque
...
+
+	recStatus = r.checkMattermost(mattermost, dbConfig, fileStoreConfig, &status, reqLogger)
+	if recStatus.Error != nil {
+		if recStatus.Status {
+			time.Sleep(5 * time.Second)
+		}
+		r.updateStatusReconcilingAndLogError(mattermost, status, reqLogger, recStatus.Error)
+		return reconcile.Result{}, recStatus.Error
 	}
```
Szymongib, line 159: We do not need to sleep here, we can requeue the resource after some time instead:
```suggestion
	recStatus, err := r.checkMattermost(mattermost, dbConfig, fileStoreConfig, &status, reqLogger)
	if err != nil {
		r.updateStatusReconcilingAndLogError(mattermost, status, reqLogger, recStatus.Error)
		return reconcile.Result{}, recStatus.Error
	}
	if !recStatus.ResourceReady {
	    // TODO: add some log here
	    return ctrl.Result{RequeueAfter: 10*time.Second}, nil // TODO: extract duration as constant
	}
```

controllers/mattermost/mattermost/mattermost.go
```diff
@@ -20,55 +21,83 @@ import (
...
+		recStatus.Error = err.Error
+		return recStatus
 	}
 
-	return nil
+	return recStatus
 }
 
-func (r *MattermostReconciler) checkLicence(mattermost *mmv1beta.Mattermost) error {
+func (r *MattermostReconciler) checkLicence(mattermost *mmv1beta.Mattermost) reconcileStatus {
```
Szymongib, line 77: I think it is enough if we return `reconcileStatus` only from `checkMattermostDeployment` and `checkMattermost`. Lets keep the remaining functions simple as long as we do not need it.

controllers/mattermost/mattermost/mattermost.go
```diff
@@ -85,116 +114,169 @@ func (r *MattermostReconciler) checkMattermostService(
...
+	return recStatus
 }
 
 func (r *MattermostReconciler) checkMattermostDeployment(
 	mattermost *mmv1beta.Mattermost,
 	dbConfig mattermostApp.DatabaseConfig,
 	fileStoreInfo *mattermostApp.FileStoreInfo,
 	status *mmv1beta.MattermostStatus,
-	reqLogger logr.Logger) error {
+	reqLogger logr.Logger) reconcileStatus {
```
Szymongib, line 274: Let's keep the error here separate, as I mentioned in the above comment:
```suggestion
	reqLogger logr.Logger) (reconcileStatus, error) {
```

Szymongib (commented):
Thanks for the PR! 

I do not think we need to go that far with the new `reconcileStatus`. Let's keep it separate from error and use it only in places that we need it.

imskr:
Thanks for reviewing @Szymongib I will make the necessary changes :)

controllers/mattermost/mattermost/mattermost.go
imskr: Okay Thanks. We also need it to return from `updateMattermostDeployment`

imskr:
I've added the changes @Szymongib

controllers/mattermost/mattermost/controller.go
```diff
@@ -144,12 +145,17 @@ func (r *MattermostReconciler) Reconcile(ctx context.Context, request ctrl.Reque
...
 
-	err = r.checkMattermost(mattermost, dbConfig, fileStoreConfig, &status, reqLogger)
+	recStatus, err := r.checkMattermost(mattermost, dbConfig, fileStoreConfig, &status, reqLogger)
 	if err != nil {
 		r.updateStatusReconcilingAndLogError(mattermost, status, reqLogger, err)
 		return reconcile.Result{}, err
 	}
 
+	if !recStatus.ResourcesReady {
+		reqLogger.Info("Mattermost resources not ready, dealying for 10 seconds!")
```
gabrieljackson, line 155:
```suggestion
		reqLogger.Info("Mattermost resources not ready, delaying for 10 seconds!")
```

controllers/mattermost/mattermost/mattermost.go
```diff
@@ -334,12 +350,13 @@ func (r *MattermostReconciler) updateMattermostDeployment(
 		defer r.cleanupUpdateJob(job, reqLogger)
 	}
 	if err != nil {
-		return err
+		recStatus.ResourcesReady = false
+		return recStatus, err
```
gabrieljackson, line 354: I believe we have an issue here. Although we are setting `ResourcesReady` to false and passing it back up with `recStatus`, we are also still passing up the error value. If I am not mistaken that error will make it all the way back to the `Reconcile` method where it will always be checked first meaning `recStatus` goes unused.

gabrieljackson (changes requested):
Looks like `ResourcesReady` may not be able to be properly used as is. More details below.

controllers/mattermost/mattermost/mattermost.go
Szymongib: I agree with Gabe, we need to adjust the logic here. It looks like `checkUpdateJob` will only return non `nil` job if it is completed, this should give us information on whether we set the `ResourcesReady` to true or false.

We should probably add some test to verify it.

controllers/mattermost/mattermost/controller.go
imskr: Thanks

controllers/mattermost/mattermost/mattermost.go
Szymongib: Yes I think that would make sense, however we would need to set it to `true` not `false` in that case, as `if job != nil` it means it completed successfully 🙂

imskr: cool thanks

controllers/mattermost/mattermost/mattermost.go
```diff
@@ -22,42 +22,49 @@ import (
...
+	reqLogger logr.Logger) (reconcileStatus, error) {
 	reqLogger = reqLogger.WithValues("Reconcile", "mattermost")
+	recStatus := reconcileStatus{
+		ResourcesReady: true,
+	}
 
 	err := r.checkLicence(mattermost)
 	if err != nil {
-		return errors.Wrap(err, "failed to check mattermost license secret.")
+		return recStatus, errors.Wrap(err, "failed to check mattermost license secret.")
```
Szymongib, line 42: nit: when we are returning the error I would suggest returning zero value of the struct, just to be transparent that it should not be used by the called in this case, otherwise it may be confusing if the `err != nil` but `ResourceReady == true`.
```suggestion
		return reconcileStatus{}, errors.Wrap(err, "failed to check mattermost license secret.")
```

Szymongib (commented):
Looks good!

Just a small suggestion regarding return statements. I also think we should add a small unit test for the new functionality.

controllers/mattermost/mattermost/mattermost.go
imskr: Yeah right thanks. I will return the `zero` value struct.

controllers/mattermost/mattermost/mattermost.go
```diff
@@ -332,16 +347,18 @@ func (r *MattermostReconciler) updateMattermostDeployment(
...
 	job, err := r.checkUpdateJob(mattermost, mattermost.Namespace, desired, reqLogger)
 	if job != nil {
+		recStatus.ResourcesReady = true
 		// Job is done, need to cleanup
 		defer r.cleanupUpdateJob(job, reqLogger)
 	}
 	if err != nil {
-		return err
+		recStatus.ResourcesReady = false
+		return recStatus, err
```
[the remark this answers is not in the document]
imskr: Thanks I have added a check to assert `false`

controllers/mattermost/mattermost/mattermost.go
gabrieljackson: Sorry, it looks like I can be a little more clear on what I was originally trying to point out. So, if I am following this correctly then `checkUpdateJob` will return an error unless the whole thing completes successfully. Now, the code I linked here returns both the `checkUpdateJob` err value as well as recStatus with `ResourcesReady=false`.

The issue I see here is that the `Reconcile` method above has this:

```
	recStatus, err := r.checkMattermost(mattermost, dbConfig, fileStoreConfig, &status, reqLogger)
	if err != nil {
		r.updateStatusReconcilingAndLogError(mattermost, status, reqLogger, err)
	}

	if !recStatus.ResourcesReady {
		reqLogger.Info("Mattermost resources not ready, delaying for 10 seconds!")
		return ctrl.Result{RequeueAfter: resourcesReadyDelay}, nil
	}
```
	
Because an error is always returned it will hit that logic fist and never get to the `ResourcesReady` check. This means we will still run into the issue noted in the linked issue.

> From reconciler perspective, when the job is still running the error is returned to requeue reconciliation request. If error occur enough number of times, in rare cases this may significantly delay rolling out the deployment due to error reconciliation back off.

Szymon, let me know if I am missing something here or looking at this the wrong way.

Szymongib: I see what you are saying, yes I think you are right, we would need to adjust this part https://github.com/mattermost/mattermost-operator/blob/master/controllers/mattermost/mattermost/mattermost.go#L378-L396 and probably here https://github.com/mattermost/mattermost-operator/blob/master/controllers/mattermost/mattermost/mattermost.go#L362 to not return errors when job was created but not finished but instead set `ResourceReady` to `false`.

That is way I think we need some tests for that, we could add it here https://github.com/mattermost/mattermost-operator/blob/master/controllers/mattermost/mattermost/mattermost_test.go#L283 with something simple such as:
```go
// Check deployment - update job is not completed, therefore resource is not ready
recResult, err = reconciler.checkMattermostDeployment(mm, dbInfo, fileStoreInfo, currentMMStatus, logger)
require.NoError(t, err)
assert.False(t, recResult.ResourceReady)
```

imskr: Yeah understood we certainly do not have to return error as mentioned by @Szymongib. Thanks for reviewing this @gabrieljackson @Szymongib :)

gabrieljackson:
Hey @imskr let us know if you need any extra support on this.

controllers/mattermost/mattermost/mattermost.go
```diff
@@ -350,19 +368,23 @@ func (r *MattermostReconciler) checkUpdateJob(
...
 	if err != nil {
 		if k8sErrors.IsNotFound(err) {
 			reqLogger.Info("Launching update image job")
 			if err = r.Resources.LaunchMattermostUpdateJob(mattermost, jobNamespace, baseDeployment, reqLogger); err != nil {
-				return nil, errors.Wrap(err, "Launching update image job failed")
+				return nil, reconcileStatus{}, errors.Wrap(err, "Launching update image job failed")
 			}
-			return nil, errors.New("Began update image job")
+			recStatus.ResourcesReady = false
+			return nil, recStatus, errors.Wrap(err, "failed to restart update job")
```
Szymongib, line 384: Here we actually managed to restart the job successfully, so we should not return an error, just that resource is not ready.

```suggestion
			return nil, recStatus, nil
```

controllers/mattermost/mattermost/mattermost.go
```diff
@@ -373,31 +395,35 @@ func (r *MattermostReconciler) checkUpdateJob(
...
+		err = r.Resources.RestartMattermostUpdateJob(mattermost, job, baseDeployment, reqLogger)
 		if err != nil {
-			return nil, errors.Wrap(err, "failed to restart update job")
+			recStatus.ResourcesReady = false
+			return nil, recStatus, errors.Wrap(err, "failed to restart update job")
 		}
 
-		return nil, errors.New("Restarted update image job")
+		recStatus.ResourcesReady = false
+		return nil, recStatus, errors.New("Restarted update image job")
```
Szymongib, line 409: We restarted the job successfully, so again we just want `ResourcesReady == false` and no error.

```suggestion
		return nil, recStatus, nil
```

controllers/mattermost/mattermost/mattermost.go
```diff
@@ -373,31 +395,35 @@ func (r *MattermostReconciler) checkUpdateJob(
...
 	}
 
 	if job.Status.CompletionTime == nil {
-		return nil, errors.New("update image job still running")
+		recStatus.ResourcesReady = false
+		return nil, recStatus, errors.New("update image job still running")
```
Szymongib, line 414:
```suggestion
		return nil, recStatus, nil
```

controllers/mattermost/mattermost/mattermost_test.go
```diff
@@ -265,10 +266,12 @@ func TestCheckMattermost(t *testing.T) {
 
 		err = reconciler.Client.Update(context.TODO(), modified)
 		require.NoError(t, err)
+		recStatus, err = reconciler.checkMattermostDeployment(mm, dbInfo, fileStoreInfo, currentMMStatus, logger)
```
Szymongib, line 269: Lets move it after we assert that job was created, but before we update it with proper status (around line 286). We also should assert it does not error and the `ResourceReady` is `false`.

controllers/mattermost/mattermost/mattermost_test.go
```diff
@@ -265,10 +266,12 @@ func TestCheckMattermost(t *testing.T) {
...
 
 		// Update job should be launched, we expect error
-		err = reconciler.checkMattermostDeployment(mm, dbInfo, fileStoreInfo, currentMMStatus, logger)
-		require.Error(t, err)
+		recStatus, err = reconciler.checkMattermostDeployment(mm, dbInfo, fileStoreInfo, currentMMStatus, logger)
+		require.NoError(t, err)
```
Szymongib, line 273: After changes proposed above we expect no error, so we should adjust the comment.

imskr:
Thanks for reviewing @Szymongib I've added the changes :)

controllers/mattermost/mattermost/mattermost.go
```diff
@@ -381,7 +381,8 @@ func (r *MattermostReconciler) checkUpdateJob(
 				return nil, reconcileStatus{}, errors.Wrap(err, "Launching update image job failed")
 			}
 			recStatus.ResourcesReady = false
-			return nil, recStatus, errors.Wrap(err, "failed to restart update job")
+			reqLogger.Error(err, "failed to restart update job")
```
Szymongib, line 384: Lets remove this error log, nothing actually failed here, we just did not find existing job so we started a new one.

controllers/mattermost/mattermost/mattermost.go
```diff
@@ -402,23 +403,27 @@ func (r *MattermostReconciler) checkUpdateJob(
 		err = r.Resources.RestartMattermostUpdateJob(mattermost, job, baseDeployment, reqLogger)
 		if err != nil {
 			recStatus.ResourcesReady = false
-			return nil, recStatus, errors.Wrap(err, "failed to restart update job")
+			reqLogger.Error(err, "failed to restart update job")
+			return nil, recStatus, nil
```
Szymongib, line 407: In this case, we should return an error as the restart of the update job failed.

controllers/mattermost/mattermost/mattermost.go
```diff
@@ -402,23 +403,27 @@ func (r *MattermostReconciler) checkUpdateJob(
...
+		return nil, recStatus, nil
 	}
 
 	// Job is completed, can check completion status
 
 	if job.Status.Failed > 0 {
 		recStatus.ResourcesReady = false
-		return job, recStatus, errors.New("update image job failed")
+		reqLogger.Info("update image job failed")
+		return job, recStatus, nil
```
Szymongib, line 426: Similarly here, if the job failed, we should report it as an error.

controllers/mattermost/mattermost/mattermost_test.go
```diff
@@ -292,6 +295,7 @@ func TestCheckMattermost(t *testing.T) {
 		}
 		err = reconciler.Client.Update(context.TODO(), job)
 		require.NoError(t, err)
+		assert.False(t, recStatus.ResourcesReady)
```
Szymongib, line 298: Here we are not checking anything as we already checked this `recStatus`. This check should be after next call to `checkMattermostDeployment` and we should expect `True`.

controllers/mattermost/mattermost/mattermost.go
imskr: ok

controllers/mattermost/mattermost/mattermost_test.go
imskr: okay

Szymongib (approved):
Looks good! Thanks for making all the changes 🙂

gabrieljackson (approved):
Thanks again for the contribution! Looks good. 👍

PR #291, opened 2022-07-08
Support parallel reconciliation

Szymongib:
#### Summary

Support parallel reconciliation of Mattermost custom resources.

Given that we have a custom mechanism for rate-limiting reconciling Mattermost instances, this needs to be done carefully so that we do not exceed the limit due to a race condition between different workers.

#### Release Note

```release-note
Support parallel reconciliation
```


gabrieljackson (approved):
Awesome!

PR #292, opened 2022-07-19
Introduce the 'ready' state to signal instance readiness

Szymongib:
#### Summary

An instance is marked as `ready` when at least one pod is updated and all other resources are ready as well.

#### Release Note

```release-note
Introduce the 'ready' state to signal instance readiness
```


gabrieljackson (approved):
Looks great!

Issue #244, opened 2021-08-26
Changing service type for Mattermost managed by Operator requires deleting service


Szymongib:
#### Summary
When changing the service type used by Mattermost with `useServiceLoadBalancer` either `true -> false` or `false -> true` the manual intervention of deleting the Service is required as some immutable fields change.

Mattermost Operator should be able to handle it either by doing smart update omitting immutable fields (if possible) or recreating the Service automatically.

#### Steps to reproduce

- Create any Mattermost instance with `useServiceLoadBalancer` set to `true`.
- Wait until Service is created.
- Edit Mattermost manifest setting `useServiceLoadBalancer` to `false`.

The reverse change (from `false` to `true`) also produces the error and should be handled properly.

#### Expected behavior
Service is updated and Mattermost reaches a `stable` state.

#### Observed behavior (that appears unintentional)
Mattermost is stuck in `reconciling` state and Mattermost Operator logs errors.

#### Possible fixes
Perform Service update omitting immutable fields (if possible) or recreate the Service automatically.

MarkAndersonTrocme:
@gabrieljackson I'll take a look at this one

gabrieljackson:
Awesome! Thanks for picking it up.

PR #295, opened 2022-07-22
Fixed changing service type between headless service and LoadBalancer

MarkAndersonTrocme:
#### Summary

When the service type changes between headless and loadbalancer, the `ClusterIP` field is immutable. This requires the service to be deleted and re-created.
#### Ticket Link

Fixes https://github.com/mattermost/mattermost-operator/issues/244

#### Release Note

```release-note
Fixed changing service type between headless service and LoadBalancer
```

### DIFF
--- a/controllers/mattermost/mattermost/mattermost.go
+++ b/controllers/mattermost/mattermost/mattermost.go
@@ -104,6 +104,7 @@ func (r *MattermostReconciler) checkMattermostService(
 	}
 
 	if current.Spec.Type != desired.Spec.Type && current.Spec.Type != "" {
+		reqLogger.Info("Recreate service due to service type change", "name", current.Name)
 		err := r.Resources.DeleteService(types.NamespacedName{Namespace: current.Namespace, Name: current.Name}, reqLogger)
 		if err != nil {
 			return errors.Wrap(err, "failed to delete service")
--- a/controllers/mattermost/mattermost/mattermost.go
+++ b/controllers/mattermost/mattermost/mattermost.go
@@ -104,11 +104,12 @@ func (r *MattermostReconciler) checkMattermostService(
 	}
 
 	if current.Spec.Type != desired.Spec.Type && current.Spec.Type != "" {
-		reqLogger.Info("Recreate service due to service type change", "name", current.Name)
 		err := r.Resources.DeleteService(types.NamespacedName{Namespace: current.Namespace, Name: current.Name}, reqLogger)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


controllers/mattermost/mattermost/mattermost.go
```diff
@@ -103,6 +103,13 @@ func (r *MattermostReconciler) checkMattermostService(
 		return err
 	}
 
+	if current.Spec.Type != desired.Spec.Type && current.Spec.Type != "" {
+		err := r.Resources.DeleteService(types.NamespacedName{Namespace: current.Namespace, Name: current.Name}, reqLogger)
```
Szymongib, line 107: It would be good to add some log here that we are going to recreate service since the type have changed.

controllers/mattermost/mattermost/mattermost.go
```diff
@@ -103,6 +103,13 @@ func (r *MattermostReconciler) checkMattermostService(
...
+		if err != nil {
+			return errors.Wrap(err, "failed to delete service")
+		}
+	}
```
Szymongib, line 111: We are trying to update the service after the deletion, which will always fail. This probably works due to the CR being requeued for reconciliation but produces and error log for the behaviour that is expected.

I think we could just attempt to recreate the Service and return early, since we do not need to copy auto assigned fields as the Service is recreated anyway.
```suggestion
	if current.Spec.Type != desired.Spec.Type && current.Spec.Type != "" {
		err := r.Resources.DeleteService(types.NamespacedName{Namespace: current.Namespace, Name: current.Name}, reqLogger)
		if err != nil {
			return errors.Wrap(err, "failed to delete service")
		}
		return r.Resources.Create(mattermost, desired, reqLogger)
	}
```

Szymongib (commented):
Thanks a lot for contribution @MarkAndersonTrocme, looks good!

I have added some small suggestions. Also, I think that it would be good to add this to tests, you can take a look [here](https://github.com/mattermost/mattermost-operator/blob/master/controllers/mattermost/mattermost/mattermost_test.go#L62), perhaps adding new test (something like `recreate service on type change`) after this one wouldn't be to complicated.

controllers/mattermost/mattermost/mattermost.go
MarkAndersonTrocme: Done!

Szymongib (approved):
Looks good! Thanks!

gabrieljackson (approved):
Looks great! Thanks for the contribution.

Issue #286, opened 2022-05-31
mattermost-operator does not provide appropriate configuration for securityContext in mattermost deploymets.


frjaraur:
#### Summary
Although mattermost-operator deployment via chart allows securityContext configuration, it is not applied to upstream mattermost installations.
I reviewed CRD definition and it does not appear a securityContext key to configure this secure behavior, therefore mattermost installations fail on secure environments (Pod Security Policies enabled not allowing root containers/specify runAs user). This just appears for special initContainers.

#### Steps to reproduce

Deploy the mattermost-operator and create a new mattermost instance.

#### Expected behavior
I expect to be able to specify the securityContext key for upstream mattermost instances created using mattermost-operator.

#### Observed behavior (that appears unintentional)
CreateContainerConfigError arraise as soon as you create a new mattermost instance due to not being able to especify RunAs user key and image defined user is not represented by its ID.

Editing created deployment manually works just by adding requiredsecurityContext configuration, but first it is required to change ownerReferences's controller key to false to disable the management of this resource by the mattermost-operator.

#### Possible fixes
Add securityContext option for being able to configure Kubernetes environment security for both, mattermost-operator and upstream created mattermost instances using it.

frjaraur:
Sorry, I forgot that neither mattermost-operator nor mattermost instances provide securityContext configurations, I did myself for mattermost operator but it should be added also to mattermost deployments.

Szymongib:
Hey @frjaraur, thanks for reporting, this came up in the past, and I think we should add it to Mattermost CR. 

For now, there exist a workaround with use of `resourcePatch`, you can find the details here https://community.mattermost.com/core/pl/dricyco3sbn85kacom55mq9zey

Szymongib:
From the linked thread, you can set it by adding to Mattermost resource:
```yaml
  resourcePatch:
    deployment:
      patch: '[{"op":"add","path":"/spec/template/spec/initContainers/0/securityContext","value":{"runAsUser": 1337}}]'
```

Operator needs to be at in version 1.17+ and CRDs need to match this version.

frjaraur:
Thanks, but it seems it will work for initial containers only or did you just post an example?

Szymongib:
Ah, yeah, that is just example copied straight from the channel. It should be possible to adjust it according to the needs.

Szymongib:
We would like to add a native support for `securityContext` therefore reopening this.

Issue #290, opened 2022-06-29
Allow overrides to update-check job spec


mjnagel:
#### Summary

From a quick scan through the code I don't believe there is currently any way to override parts of the spec for the update-check job that runs with version updates. It may be desirable to override various parts of this job, for example:
- securityContext: Enforce best practices for security such as runAsNonRoot, dropping all capabilities, etc
- labels/annotations: To better group things together for ownership and provide for integration with externally managed resources (i.e. networkPolicy pod selectors)
- startup command: This is a very niche use-case but when deploying in an istio-injected environment the sidecar needs to be terminated by running additional commands against its endpoint (can provide more details if needed)

Ultimate it would be great to be able to override different portions of the pod spec.

#### Possible fixes

Add a new section to the Mattermost CRD spec that allows for update check pod spec overrides. This would honestly be beneficial for the primary Mattermost pods as well, since currently you have to utilize `resourcePatch` to add labels/annotations to the pods (which is not very beginner friendly or easy to navigate, especially with multiple changes.

Proposed Change:
```yaml
spec:
  podTemplate:
    <allow anything for podspec, or allow subset including annotations/labels/securitycontext
  updateCheck:
    podTemplate:
      <allow anything for podspec, or allow subset including annotations/labels/securitycontext
```

mjnagel:
I'm happy to attempt work on these changes, wanted to open the issue to get the conversation started and see if this would be supported. The primary motivation for this is coming from my work on istio injecting Mattermost, without an ability to disable the sidecar (via labels) or terminate the sidecar (via command override) for the update-check upgrades will require manual intervention to successfully complete.

Szymongib:
Hey @mjnagel, thanks for submitting the issue.

I agree that adding a way to configure an update job would be very useful, as well as other things in the `podTemplate` section. We though about something like that in the past, however there are some drawback that make me a little hesitant for allowing overrides of the whole spec.

A big chunk of `podTemplate` is created based on other things in Mattermost CR or some Operator defaults. This poses a problem, what to do when the user provides his own template? We could just use the provided one and disregard all things that would be set by the Operator, but then user would need to specify all init containers, port, env vars etc. by hand so it seems like using Operator make little sense in such case. We could alternatively do some kind of merging, but then it is also not clear what to do if user explicitly wanted to omit some configuration set by the Operator. We might also run into some configuration edge cases given that some things in Pod Spec can be configured from other CR fields and this information may be used in other places in Operator.

[...]

PR #297, opened 2022-08-02
Add securityContext to MM Spec

mjnagel:
#### Summary

This adds a new struct and two new "options" to the MM spec:
```yaml
spec:
  podTemplate:
    securityContext: {}
    containerSecurityContext: {}
```

Currently this addition of `podTemplate` only supports `securityContext` and `containerSecurityContext`, but ideally this can be extended to add support for other portions of the pod template.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-operator/issues/286

#### Release Note

```release-note
Add support for pod and container security context to be specified for server pods
```


mjnagel:
Wanted to add `securityContext` initially since its a little more straightforward to handle than some of the other spec additions I mentioned in https://github.com/mattermost/mattermost-operator/issues/290 . Hopefully the spec additions make sense - I'd like to build upon this struct in future MRs.

mjnagel:
@Szymongib / @gabrieljackson would either of you be able to take a look at this?

Szymongib:
Hey @mjnagel, thanks for the PR.

Sorry I missed it, looks like we did not get Mattermost notifications for it.

I will have a look this week.

apis/mattermost/v1beta1/mattermost_types.go
```diff
@@ -204,6 +208,16 @@ type Probes struct {
...
 }
 
+// PodTemplate defines configuration for the template for Mattermost pods.
+type PodTemplate struct {
+	// Defines the security context for the Mattermost app server pods.
+	// +optional
+	SecurityContext v1.PodSecurityContext `json:"securityContext,omitempty"`
+	// Defines the security context for the Mattermost app server container.
+	// +optional
+	ContainerSecurityContext v1.SecurityContext `json:"containerSecurityContext,omitempty"`
```
Szymongib, line 218: I think we should use pointers for those to clearly indicate the absence, otherwise this will be automatically initialized to an empty struct, bringing all the zero values along, which in turn might result in some unintended behaviour.

Szymongib (commented):
Thanks a lot for PR!
Looks good, only one minor comment regarding the pointers.

apis/mattermost/v1beta1/mattermost_types.go
mjnagel: Updated, let me know if there's anything else (or if I changed things incorrectly 😁 ).

Szymongib (approved):
LGTM! Thanks!

gabrieljackson (approved):
Looks great! Thanks for the contribution.

PR #300, opened 2022-08-23
Avoid rewriting command for injected sidecars

TMaYaD:
#### Summary
We are injecting cloudsql-proxy sidecar via a `resourcePatch`. However the operator is changing the command in that container as well to `mattermost version`. This patch will avoid that by matching the container name.

#### Release Note

```release-note
Fix update-check overriding commands in all containers.
```

### DIFF
--- a/pkg/resources/update_job.go
+++ b/pkg/resources/update_job.go
@@ -4,6 +4,7 @@ import (
 	"context"
 
 	"github.com/go-logr/logr"
+	"github.com/mattermost/mattermost-operator/apis/mattermost/v1beta1"
 	mattermostApp "github.com/mattermost/mattermost-operator/pkg/mattermost"
 	"github.com/pkg/errors"
 	appsv1 "k8s.io/api/apps/v1"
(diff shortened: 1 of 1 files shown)


TMaYaD:
Instead of rewriting the command, Setting args to `version` might be a better choice.
I can think of scenarios where a wrapper like [kubexit](https://github.com/karlkfi/kubexit) could be injected into mattermost container to manage sidecars.

Using `args` instead of `command` would allow keeping the wrapper intact and still override the behaviour.

pkg/resources/update_job.go
```diff
@@ -103,7 +103,9 @@ func PrepareMattermostJobTemplate(name, namespace string, baseDeployment *appsv1
 	// Override values for job-specific behavior.
 	job.Spec.Template.Spec.RestartPolicy = corev1.RestartPolicyNever
 	for i := range job.Spec.Template.Spec.Containers {
-		job.Spec.Template.Spec.Containers[i].Command = []string{"mattermost", "version"}
+		if job.Spec.Template.Spec.Containers[i].Name == "mattermost" {
+			job.Spec.Template.Spec.Containers[i].Command = []string{"mattermost", "version"}
+		}
 	}
```
mjnagel, line 109: You should be able to use [this function](https://github.com/mattermost/mattermost-operator/blob/master/apis/mattermost/v1beta1/mattermost_utils.go#L152) to grab the `mattermost` container and likely remove the `for` loop entirely here since the intended target of this command override should just be the `mattermost` container.

TMaYaD: That wouldn't work out of the box because the container being returned is a copy. I can probably modify the function to pass a reference instead, but that might have other unintended side effects.

mjnagel: Ah yep good point. Probably best to not try to modify that and just go with the way you approached it already. I'm interested in this fix as well because we're using istio injection (side cars).

mjnagel (commented):
https://github.com/mattermost/mattermost-operator/pull/300/files#diff-3ea61f19d355d951d2aa6a84e23e64bc77116044de834d8ce764e7042f3016a6R99

There's a similar section here (lines 99-102) that nulls out the liveness/readiness probes. I assume an identical fix would be good, maybe even combining these in the same for loop to simplify things.

TMaYaD:
The two loops were there from before the PR. I can merge them, no problem.

mjnagel:
> The two loops were there from before the PR. I can merge them, no problem.

Tracking that they were both there. My comment was mostly that you'd probably want the same effect on the health checks - only nil out the update container, and ignore any extras. So you'd want to move that inside of your conditional container name check. Definitely not something broke, but seems like a good addition that falls in line with this fix too.

TMaYaD:
Ah, okay. Got it.

mjnagel (approved):
LGTM!

Szymongib (approved):
LGTM, thanks!

gabrieljackson (approved):
LGTM. Thanks!

PR #301, opened 2022-08-25
Add support for labels/annotations and disabling on update job

mjnagel:
#### Summary

This adds a new struct and three new "options" to the MM spec:

```yaml
spec:
  updateJob:
    extraLabels: {}
    extraAnnotations: {}
    disabled: (bool)
```

The annotations/labels will be added to the update job pod, but will not override the default `app` label set on the job pod.

`disabled` will control the creation/use of the update job, defaulting to the same behavior as previously.

Also adds tests for each of these new pieces.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-operator/issues/290

#### Release Note

```release-note
Add support for labels and annotations to be specified for update job pod
Add ability to disable the update job pod for advanced configuration
```


mjnagel:
@Szymongib leaving this in draft right now because I have two questions regarding it...
1. Clusterinstallation uses the same functions for the update job. Since these changes require adding a new parameter to pass these update functions part of the MM spec I needed to address the function calls from clusterinstallation code. I'm not sure if the way I approached it is the best, but it effectively passes in an empty struct to fulfill the required params. Happy to alter this if there's a better way to approach - only other thing I could think of was make separate functions for each which isn't very DRY.
2. Part of the issue I opened (https://github.com/mattermost/mattermost-operator/issues/290) was the goal of allowing a modification/extension of the command this job runs. My use case would be to run an istio sidecar termination script after the main `mattermost version` command has run. I imagine others might have similar desire to run something after the upgrade, but tied to the lifecycle of the update job. If this is something you'd be open to supporting what would be the best way to approach it? Allow a full override of the command (`spec.updateJob.command`), which would be the most flexible but open end users to deleting the main upgrade functionality? Alternatively we could allow for something that would be appended to/run after the default command (`spec.updateJob.postUpgradeCommand`) which would fit most needs I could think of?

Szymongib:
Thanks for another PR @mjnagel.

1. I think passing empty structs as arguments for the ClusterInstallation code would be fine since this will not alter the behavior.
2. I am still not sure what is the best way to handle overriding the update job command. I think that the simplest approach would be just to give the ability to disable it. It does not have any special behavior rather than failing if the image is incorrect so if it is disabled, new pods just won't come up therefore update will fail but the old pods should remain running.

mjnagel:
@Szymongib appreciate the feedback, I believe I've implemented a disable value in a simple way that should work (and added some tests around all of this new functionality). Let me know if any changes are needed, hopefully this is good to go 😄 

---

I do want to add a comment to one thing you mentioned...

> It does not have any special behavior rather than failing if the image is incorrect

In our usage the update job has been extremely helpful for the DB migrations - if I'm not mistaken it will perform them on startup (and I saw a comment in the code to that effect as well). It's not super special (no magical code doing this that we couldn't implement), but it is nice to have this all handled by the operator and tied into the lifecycle in the right order.

I'm good with the disabled solution for now since it provides a path forward with our primary use case for istio injection, but might be something to give some more thought to in the future.

Szymongib:
> In our usage the update job has been extremely helpful for the DB migrations

DB migrations should run auto automatically when the new instance start. That what I was referring to about job not really doing any magic, sorry, I could have been a bit cleaner on that.

controllers/mattermost/mattermost/mattermost.go
```diff
@@ -349,13 +349,19 @@ func (r *MattermostReconciler) updateMattermostDeployment(
 	}
 
 	// Image is not the same
+
+	reqLogger.Info("Current image is not the same as the requested, will upgrade the Mattermost installation")
+
+	if mattermost.Spec.UpdateJob.Disabled {
+		// Update job is disabled
+		return recStatus, r.Resources.Update(current, desired, reqLogger)
```
Szymongib, line 357: I think it would be good to add a Debug log here.

pkg/resources/update_job.go
```diff
@@ -70,9 +73,22 @@ func (r *ResourceHelper) FetchMattermostUpdateJob(namespace string) (*batchv1.Jo
 	return job, err
 }
 
-func PrepareMattermostJobTemplate(name, namespace string, baseDeployment *appsv1.Deployment) *batchv1.Job {
+func PrepareMattermostJobTemplate(name, namespace string, baseDeployment *appsv1.Deployment, updateJobSpec *mmv1beta.UpdateJob) *batchv1.Job {
 	backoffLimit := int32(10)
 
+	podAnnotations := map[string]string{}
+	// Set user specified annotations
+	if updateJobSpec.ExtraAnnotations != nil {
```
Szymongib, line 81: I think it would be good to do nil check here for `UpdateJob` then we can pass nil for ClusterInstallation.

apis/mattermost/v1beta1/mattermost_types.go
```diff
@@ -117,6 +117,10 @@ type MattermostSpec struct {
 	// +optional
 	PodTemplate PodTemplate `json:"podTemplate,omitempty"`
 
+	// UpdateJob defines configuration for the template for the update job.
+	// +optional
+	UpdateJob UpdateJob `json:"updateJob,omitempty"`
```
Szymongib, line 122: Lets make it a pointer to clearly indicate it is optional and avoid any unexpected issues in yaml because of that. I think we should make a `PodTemplate` a pointer as well it is generally a better practice since if we add some non-pointer field to the struct it may show up in yaml. No need to change it in this PR, just realized that we have missed it last time.

Szymongib (commented):
Looks good, just few small comments.

apis/mattermost/v1beta1/mattermost_types.go
mjnagel: I believe I resolved this and the other comments - I'll make a separate PR to change `PodTemplate` to a pointer just to keep things smaller and cleanly separate the changes.

mjnagel: Resolving `podTemplate` comments with https://github.com/mattermost/mattermost-operator/pull/302 - wouldn't be surprised if there will be conflicts between the two once one is merged so I'll keep an eye on these and rebase as needed.

controllers/mattermost/mattermost/mattermost_test.go
```diff
@@ -389,6 +389,148 @@ func TestCheckMattermost(t *testing.T) {
...
+
+		// Update job should NOT be launched, we do not expect error
+		recStatus, err = reconciler.checkMattermostDeployment(mm, dbInfo, fileStoreInfo, currentMMStatus, logger)
+		require.NoError(t, err)
+		assert.True(t, recStatus.ResourcesReady)
+
+		// Assert the job was NOT created
+		job := &batchv1.Job{}
+		err = reconciler.Client.Get(context.TODO(), types.NamespacedName{Name: updateJobName, Namespace: mmNamespace}, job)
+		require.Error(t, err)
```
Szymongib, line 529: We can additionally check if the error `IsNotFound`:
```suggestion
		require.Error(t, err)
		require.True(t, k8sErrors.IsNotFound(err), "expected not found error when getting job")
```

controllers/mattermost/clusterinstallation/mattermost.go
```diff
@@ -235,7 +236,8 @@ func (r *ClusterInstallationReconciler) checkMattermostDeployment(mattermost *ma
 }
 
 func (r *ClusterInstallationReconciler) checkMattermostDBSetupJob(mattermost *mattermostv1alpha1.ClusterInstallation, deployment *appsv1.Deployment, reqLogger logr.Logger) error {
-	desiredJob := resources.PrepareMattermostJobTemplate(mattermostApp.SetupJobName, mattermost.Namespace, deployment)
+	// Create an empty v1beta update job struct to support update job function changes
```
Szymongib, line 239: Since we are passing `nil` now, I think we can remove those comments.

Szymongib (approved):
Hey @mjnagel, sorry for delay I was off last week. Looks good, just to more tiny nit picks 😉

apis/mattermost/v1beta1/mattermost_types.go
Szymongib: Thanks for changing that as well, separate PR works just fine.

controllers/mattermost/mattermost/mattermost_test.go
mjnagel: @Szymongib not a problem! Should be good to go?

controllers/mattermost/clusterinstallation/mattermost.go
```diff
@@ -397,7 +398,8 @@ func (r *ClusterInstallationReconciler) checkUpdateJob(
 		if k8sErrors.IsNotFound(err) {
 			// Job is not running, let's launch
 			reqLogger.Info("Launching update image job")
-			if err = r.Resources.LaunchMattermostUpdateJob(mattermost, mattermost.Namespace, desired, reqLogger); err != nil {
+			// Create an empty v1beta update job struct to support update job function changes
```
Szymongib, line 401: I think there are 2 more of those 😉 
```suggestion
```

controllers/mattermost/clusterinstallation/mattermost.go
```diff
@@ -419,7 +421,8 @@ func (r *ClusterInstallationReconciler) checkUpdateJob(
 	}
 	if !isSameImage {
 		reqLogger.Info("Mattermost image changed, restarting update job")
-		err := r.Resources.RestartMattermostUpdateJob(mattermost, job, desired, reqLogger)
+		// Create an empty v1beta update job struct to support update job function changes
```
Szymongib, line 424:
```suggestion
```

Szymongib (approved):

gabrieljackson:
Looks like there is a merge conflict as expected from the related PR. If you could resolve that we will get this merged asap. Thanks!

mjnagel:
@gabrieljackson / @Szymongib should be good to go? Rebased and fixed those comments - appreciate the feedback and assistance 😄

gabrieljackson (approved):
LGTM!

gabrieljackson:
Going to request one more review from Szymon to look over the final code.

Szymongib (approved):
LGTM, thanks!

PR #304, opened 2022-09-15
Fix replicas status reported in Mattermost CR

Szymongib:
#### Summary

Checking ready replicas in Mattermost Operator is not working as intended as it considers old pods that were not yet terminated as `UpdatedReplicas` if the changes to manifest were something else than the Mattermost version. This causes the Operator to improperly report the `ready` state as it relies on `UpdatedReplicas` in the status.

For this check, we do not need to check all pods one by one but rather we can rely on the status of ReplicaSet which already provides this information.

Old `CheckPodsRollOut` is left out only for the purpose of not changing the migration logic that depends on it. After it is removed it can be cleaned up as well.

Additionally, this fixes a small issue that Mattermost CR might have not been set to a `reconciling` state if the initial update failed and the update job has already started.

#### Release Note

```release-note
Fix replicas status reported in Mattermost CR
```

### DIFF
--- a/controllers/mattermost/clusterinstallation/controller_test.go
+++ b/controllers/mattermost/clusterinstallation/controller_test.go
@@ -1,7 +1,6 @@
 package clusterinstallation
 
 import (
-	"fmt"
 	"testing"
 	"time"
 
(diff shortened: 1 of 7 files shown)


gabrieljackson (approved):
Looks great!

Issue #255, opened 2021-10-18
Local File store


felix-dpg:
#### Summary
Why in the CRD definition is not present Local File store?  Its not possible to deploy mattermost in k8s using the operator (or the helm chart) with local file store? 

In the docs say that is possible to use it: https://docs.mattermost.com/configure/configuration-settings.html#file-storage-system

#### Expected behavior
Use Local File store in the mattermost deployed in k8s.

Szymongib:
In theory, it would be possible to configure local file store with environment variables. This is not possible with CRD tho, as containers file system is transient therefore unless you mount a persistent volume you would lose all your data whenever containers restart.

felix-dpg:
Yes, I will mount a persistent volume, so, please, where in the docs talk about local file store with env variable?

Szymongib:
We only recently added support for volumes and using external file store is more aligned with cloud application. Also as you mentioned there is always option for deploying Minio so I suppose there was never real need for it.

gabrieljackson:
Due to a few recent requests for this functionality in other communication channels we are opening this up as a help wanted request.

The task would be to provide a clean configuration option for a local filestore type which would be clearly marked as not supported for production environments.

mjnagel:
@Szymongib / @gabrieljackson If you want to assign me to this, I've got a PR that hopefully resolves - https://github.com/mattermost/mattermost-operator/pull/305 - happy to make any modifications needed to get it ready.

PR #305, opened 2022-09-15
Add support for local/PVC file storage

mjnagel:
#### Summary

This PR adds support for local file storage with PVC backing. Local storage must be explicitly enabled, with the option to configure the PVC size:

```yaml
spec:
  fileStore:
    local:
      enabled: true
      storageSize: "10Gi" # Optional, defaults to 50Gi same as Minio
```

No other PVC configuration is exposed, since this is not recommended for production nothing else should be required for a development/test install.

Several tests are added to validate PVC creation/resizing and proper deployment additions.

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-operator/issues/255

#### Release Note

```release-note
Add support for local file storage with PVC backing
```


Szymongib (commented):
Thanks for another PR @mjnagel!

While creating PVC by Operator might be useful for people to start quickly, I wonder if it would not be better to just pass PVC name as Local file-store configuration.

Also I wonder how this works since we do not specify the `storageClass` so from my understanding the dynamic creation of PV won't happen therefore not sure if PVC will bind to anything 🤔  However, I did not have a chance to work with volumes much so might be missing something.

Szymongib:
Thanks for clarification @mjnagel. I do not really have a strong opinion either way, would have to think a bit more about it.

@gabrieljackson any thoughts?

gabrieljackson:
I have been thinking about this for a few days and am honestly pretty split about what makes sense here. I definitely see the tradeoffs with these two options which is giving me pause. One thing I can say is that I agree that we should keep things simple and can rely on the default storage class setting if we keep the current implementation.

Regarding using a pre-created PVC, it would do allow for using multiple types of PVC on the same cluster or for using something other than the default type. I could see this being useful even though we are very serious about this not being used for production settings.

How do we feel about a hybrid approach on this? We could add a field to provide a PVC name and use that if it is filled in and if it isn't simply use the existing logic in this PR. I want to keep this simple though so this may be a bit too much. What do you think @Szymongib?

Szymongib:
Yeah, we can go with hybrid approach, the only thing that I am a bit worried is about maintenance of it, but then since we do not really have strong use case for specifying PVC yet, maybe it is worth to hold of with that feature.

gabrieljackson:
Yep, agreed. Let's proceed with the current implementation and the hybrid option can be added later if needed. I will start reviewing this shortly.

gabrieljackson (commented):
Looks good. Just one tiny request.

pkg/mattermost/mattermost_v1beta.go
```diff
@@ -179,7 +179,17 @@ func GenerateDeploymentV1Beta(mattermost *mmv1beta.Mattermost, db DatabaseConfig
 	initContainers := db.InitContainers(mattermost)
 
 	// File Store
-	envVarFileStore := fileStoreEnvVars(fileStore)
+	envVarFileStore := []corev1.EnvVar{}
+	volumes := mattermost.Spec.Volumes
+	volumeMounts := mattermost.Spec.VolumeMounts
```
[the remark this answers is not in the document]
mjnagel: Updated! Good call.

gabrieljackson (approved):
LGTM

pkg/mattermost/mattermost_v1beta.go
```diff
@@ -178,8 +178,20 @@ func GenerateDeploymentV1Beta(mattermost *mmv1beta.Mattermost, db DatabaseConfig
...
-	envVarFileStore := fileStoreEnvVars(fileStore)
+	envVarFileStore := []corev1.EnvVar{}
+	if mattermost.Spec.FileStore.Local != nil && mattermost.Spec.FileStore.Local.Enabled {
+		envVarFileStore = localFileEnvVars(mmv1beta.DefaultLocalFilePath)
+		vMount, volume := mattermostLocalFileStoreConfig(mattermost.Name)
+		volumeMounts = append(volumeMounts, vMount)
+		volumes = append(volumes, volume)
+	} else {
+		envVarFileStore = s3EnvVars(fileStore)
+	}
```
Szymongib, line 194: I feel like this part becomes a bit complicated, as well as passing empty `LocalFileStore` to the `FileStoreInfo`.

Since things are quite a bit different now for different file stores, I wonder if we should not hide it behind interface to not leak that logic in multiple places, similarly as we do for `DatabaseConfig` (https://github.com/mjnagel/mattermost-operator/blob/729370d6270fac866cd69fea844f01714d98659d/pkg/mattermost/mattermost_v1beta.go#L25-L28).

We would just have methods that return configs specific for different file stores and we avoid additional `if else` logic here.

I am thinking of something like (in pseudo code):
```
type FileStoreConfig interface {
    InitContainers(mm *v1beta1.Mattermost) []InitContainer
    Volumes(mm *v1beta1.Mattermost) ([]Volumes, []VolumeMount
    EnvVars(mm *v1beta1.Mattermost) []EnvVar
}
```
So that would mean separate `struct` for each implementation. We can perhaps use `FileStoreInfo` as a field for MinIO and S3 implementations to reuse bunch of this logic, but the Local one would have a different representation.

Let me know what do you think about this approach, and feel free to ask any follow up questions if I phrased something not clear.

mjnagel: It did seem weird to pass the empty `LocalFileStore`, but that seemed to be the best way to fit the current structure. More than happy to take a stab at restructuring a bit based on your comments, that makes a lot of sense. I'll reach out if I get lost or confused on anything.

Szymongib: Awesome, thanks a lot!

mjnagel: Sorry for the delay on this, been caught up with some other things and haven't gotten back to these changes. I started the restructure but got a little lost - will try and get back to working on this and reach out with questions.

mjnagel: @Szymongib updated! It was a lot more changes than I initially thought but I think follows the spirit of what you were asking for? Let me know if any other changes are needed.

controllers/mattermost/mattermost/file_store.go
```diff
@@ -35,7 +41,56 @@ func (r *MattermostReconciler) checkExternalFileStore(mattermost *mmv1beta.Matte
 	return mattermostApp.NewExternalFileStoreInfo(mattermost, *secret)
 }
 
-func (r *MattermostReconciler) checkOperatorManagedMinio(mattermost *mmv1beta.Mattermost, reqLogger logr.Logger) (*mattermostApp.FileStoreInfo, error) {
+func (r *MattermostReconciler) checkLocalFileStore(mattermost *mmv1beta.Mattermost, reqLogger logr.Logger) (mattermostApp.FileStoreConfig, error) {
+	// Default 1Gi volume using default storage class
```
gabrieljackson, line 45:
```suggestion
	// Default 50Gi volume using default storage class
```

controllers/mattermost/mattermost/file_store.go
```diff
@@ -35,7 +41,56 @@ func (r *MattermostReconciler) checkExternalFileStore(mattermost *mmv1beta.Matte
...
+	}
+	pvc := &corev1.PersistentVolumeClaim{
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      mattermost.Name,
+			Namespace: mattermost.Namespace,
+			Labels:    mattermost.MattermostLabels(mattermost.Name),
+		},
+		Spec: corev1.PersistentVolumeClaimSpec{
+			AccessModes: []corev1.PersistentVolumeAccessMode{
+				"ReadWriteOnce",
```
gabrieljackson, line 58: I think this AccessMode makes sense, but just wanted to check that it was an intentional choice. We can always extend the local file store configuration to add other options should the need arise, but just wanted to quickly confirm that we think this is the best suited default config.

gabrieljackson (commented):
This is looking really nice and we are getting close. Just two final comments from me.

controllers/mattermost/mattermost/file_store.go
fmartingr: Unsure about the actual need for this comment, since we could just check the constant.

fmartingr (approved):
Thank you @mjnagel ! This looks good to me and will help testing the operator locally or on my test cluster, really appreciated!

controllers/mattermost/mattermost/file_store.go
mjnagel: Removed comment.

controllers/mattermost/mattermost/file_store.go
mjnagel: Sorry for the delay on these - I missed notification of comments somehow. 

This was an intentional choice since it seems to be a common default in other apps I use. Thinking about it more, I wonder if this will cause problems with multiple replicas (we only make one PVC per MM instance, not per pod, and pods are not always on the same node). Maybe we want to make the default `ReadWriteMany` to handle that scenario? @gabrieljackson what do you think?

gabrieljackson: Yep, sounds great! I completely agree here and just wanted to check that the choice fits what you had in mind for an initial use-case. Thanks for confirming.

gabrieljackson (approved):
Looks great! Thanks for all the improvements.

Issue #306, opened 2022-09-16
Add version upgrade to e2e external test


Szymongib:
#### Summary

With have e2e test working with "external" database and filestore. It is currently very simplistic, just spawning new instance and waiting for it to be stable.

https://github.com/mattermost/mattermost-operator/blob/master/test/e2e-external/external_db_filestore_test.go#L45-L64

We can extend this test with a Mattermost version upgrade.

The test should:
- Spin up Mattermost and wait for it to become stable.
- Update Mattermost version.
- Wait for instance to be stable again.
- Assert that deployment image was successfully updated.

aerokite:
I will take this. Please assign me

gabrieljackson:
Thanks @aerokite!

aerokite:
what is expected here?

with Mattermost Version update, Server Version should be updated, right?
But, external DB & filestore will not be affected.

Szymongib:
> with Mattermost Version update, Server Version should be updated, right?
But, external DB & filestore will not be affected.

Yes, that is correct. Basically we want to extend [the test](https://github.com/mattermost/mattermost-operator/blob/master/test/e2e-external/external_db_filestore_test.go#L16) to update this version https://github.com/mattermost/mattermost-operator/blob/master/apis/mattermost/v1beta1/mattermost_types.go#L40.

aerokite:
Understood, but this test already exists [here](https://github.com/mattermost/mattermost-operator/blob/master/test/e2e/mattermost_test.go#L58).

Do you want similar test for e2e external?

Szymongib:
Yes. The other test is working with MySQL and MinIO operators and we eventually want to get rid of those, therefore we want to start migrating a bunch of those tests to new framework based on external DB and filestore.

PR #309, opened 2022-09-16
e2e test for version upgrade

aerokite:
#### Summary
Added e2e test to ensure if version upgrading works or not.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-operator/issues/306

#### Release Note

```release-note
Added e2e test to ensure if version upgrading works or not.
````

### DIFF
--- a/test/e2e-external/external_db_filestore_test.go
+++ b/test/e2e-external/external_db_filestore_test.go
@@ -7,10 +7,13 @@ import (
 
 	mmv1beta "github.com/mattermost/mattermost-operator/apis/mattermost/v1beta1"
 	ptrUtil "github.com/mattermost/mattermost-operator/pkg/utils"
+	operatortest "github.com/mattermost/mattermost-operator/test"
 	"github.com/mattermost/mattermost-operator/test/e2e"
 	"github.com/stretchr/testify/require"
+	corev1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: test/e2e-external/external_db_filestore_test.go [hunk 1]
```diff
@@ -50,10 +54,6 @@ func expectValidMattermostInstance(t *testing.T, mattermost *mmv1beta.Mattermost
 
 	err := k8sClient.Create(ctx, mattermost)
 	require.NoError(t, err)
-	defer func() {
-		err = k8sClient.Delete(context.Background(), mattermost)
-		require.NoError(t, err)
-	}()
```
Szymongib, line 56: I think we still want to cleanup the test instance after the test ends, regardless if successful or not.

Since we expect the instance to exist outside this method, we could pull out creation logic to main test function and call `defer` there.

aerokite: Yes. But figured it out that following code is actually doing the cleanup stuff.

https://github.com/mattermost/mattermost-operator/blob/master/test/e2e-external/main_test.go#L28

And I have verified. Thats why I removed this.

Szymongib: Yeah, that cleans up the whole namespace therefore the CR is cleaned up as well. Ok, lets leave it this way then.

Szymongib (approved):
LGTM, thanks 🙂

gabrieljackson (approved):
Looks great. Thanks!

Issue #307, opened 2022-09-16
Add e2e external test for sizes and resource constraints


Szymongib:
#### Summary
We can add e2e test using the "external" DB and filestore to verify that sizing and resource constraint logic works correctly.

The test can:
- Spin up Mattermost instance with specified size.
- Assert that it starts correctly and that replicas and resources are set based on size.
- Assert that size value was removed.
- Manually change resource constraints.
- Assert that resources where adjusted and instance is stable.
- Set a different size on Mattermost instance.
- Again assert that resources were adjusted properly.

aerokite:
I will take it

PR #311, opened 2022-09-17
e2e test for size & resource update

aerokite:
#### Summary
Added e2e test to see if size and resource updates work.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-operator/issues/307

#### Release Note

```release-note
Added e2e test to see if size and resource updates work.
```

### DIFF
--- a/test/e2e-external/size_test.go
+++ b/test/e2e-external/size_test.go
@@ -2,14 +2,14 @@ package e2e
 
 import (
 	"context"
+	"testing"
+
 	mmv1alpha "github.com/mattermost/mattermost-operator/apis/mattermost/v1alpha1"
 	mmv1beta "github.com/mattermost/mattermost-operator/apis/mattermost/v1beta1"
 	"github.com/mattermost/mattermost-operator/test/e2e"
 	"github.com/stretchr/testify/require"
-	appsv1 "k8s.io/api/apps/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
-	"testing"
 )
 
 // TestMattermostSize checks defaulting & updating replicas & resources from size.
@@ -44,13 +44,12 @@ func TestMattermostSize(t *testing.T) {
 	mattermost.Spec.Size = mmSize
 	instance := e2e.NewMattermostInstance(t, k8sClient, mattermost)
 
-	{
+	t.Run("checking mattermost replicas & resources", func(t *testing.T) {
 		clusterSize, err := mmv1alpha.GetClusterSize(mmSize)
 		require.NoError(t, err)
 
 		t.Log("create and waiting for Mattermost to be stable")
 		instance.CreateAndWait()
-		defer instance.Destroy()
 
 		t.Log("checking mattermost replicas & resources")
 		var newMattermost mmv1beta.Mattermost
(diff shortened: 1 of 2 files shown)


test/e2e-external/size_test.go
```diff
@@ -0,0 +1,130 @@
...
+		require.Equal(t, clusterSize.App.Replicas, *newMattermost.Spec.Replicas)
+		compareResources(t, clusterSize.App.Resources, newMattermost.Spec.Scheduling.Resources)
+
+		t.Log("checking deployment replicas & resources")
+		var mmDeployment appsv1.Deployment
+		err = k8sClient.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: namespace}, &mmDeployment)
+		require.NoError(t, err)
+		require.Equal(t, clusterSize.App.Replicas, *mmDeployment.Spec.Replicas)
+		// compare resources in deployment
+		compareResources(t, clusterSize.App.Resources, mmDeployment.Spec.Template.Spec.Containers[0].Resources)
```
Szymongib, line 73: I think we could extract this logic to some function since we repeat it after each update.

test/e2e-external/size_test.go
```diff
@@ -0,0 +1,130 @@
...
+			FileStore: mmv1beta.FileStore{
+				External: &testEnv.FileStoreConfig,
+			},
+			Database: mmv1beta.Database{
+				External: &testEnv.DBConfig,
+			},
+		},
+	}
+
+	{
```
Szymongib, line 44: Any specific reason for enclosing it in additional braces?

test/e2e-external/size_test.go
```diff
@@ -0,0 +1,130 @@
...
+		t.Logf("updating scheduling resources in mattermost object with %s\n", mmSize)
+		// update scheduling resource
+		clusterSize, err := mmv1alpha.GetClusterSize(mmSize)
+		require.NoError(t, err)
+
+		var newMattermost mmv1beta.Mattermost
+		err = k8sClient.Get(context.TODO(), mmNamespaceName, &newMattermost)
+		require.NoError(t, err)
+		newMattermost.Spec.Scheduling.Resources = clusterSize.App.Resources
+		err = k8sClient.Update(context.TODO(), &newMattermost)
```
Szymongib, line 87: Since we are using `e2e.NewMattermostInstance` we could use `instance.UpdateAndWait(...)`, but we would need to get rid of the braces.

Szymongib (approved):
Thanks for the PR! Added some suggestions.

test/e2e-external/size_test.go
aerokite: This enclosure is used to make this test easier to read. This contains code blocks that perform a certain action of the test.

test/e2e-external/size_test.go
aerokite: Right. I have missed that method.

test/e2e-external/size_test.go
aerokite: We can call a function to do these checks. But I think that will not be a good idea. You will find that, in 2nd block, we didn't check replicas. 

```go
err = k8sClient.Get(context.TODO(), types.NamespacedName{Name: name, Namespace: namespace}, &mmDeployment)
require.NoError(t, err)
compareResources(t, clusterSize.App.Resources, mmDeployment.Spec.Template.Spec.Containers[0].Resources)
```

That means we can't place a replica check in there. So either two separate functions are required, or we must check replica outside of that function.

aerokite: @Szymongib any thoughts?

test/e2e-external/size_test.go
aerokite: Resolved.

test/e2e-external/size_test.go
gabrieljackson: Instead of just braces, could you change them to Go subtests? That way we also have nice headers and test time data.

test/e2e-external/size_test.go
aerokite: Resolved

test/e2e-external/size_test.go
aerokite: That is a great idea.  Resolved with Go subtests.

PR #319, opened 2022-12-05
Add external volume filestore type

gabrieljackson:
This new filestore type allows for mounting externally managed volumes for Mattermost file storage. This could be useful in situations where S3 backed storage is not available.

Some notes:
 - Documentation for our operator filestore and database options will be revamped in a future PR. We are going to classify this option as usable, but not explicitly recommended for critical deployments. S3 will continue to be the go-to option for Mattermost filestores.
 - Depending on the container image used, we need to set `SecurityContext` as well. Specifically, this is needed when using our default Mattermost docker image for our container user to properly access the volume:

```
  Pod Template:
    Security Context:
      Fs Group:  2000
```

I am unsure if we should set this to be the default when using an external volume filestore or if it should only be documented as using other images will require different configuration. Open to thoughts on this.

Fixes https://mattermost.atlassian.net/browse/MM-48840

```release-note
Add external volume filestore type
```


pkg/mattermost/file_store.go
```diff
@@ -36,6 +36,40 @@ func (e *ExternalFileStore) Volumes(_ *mmv1beta.Mattermost) ([]corev1.Volume, []
...
+
+func (fs *ExternalVolumeFileStore) EnvVars(_ *mmv1beta.Mattermost) []corev1.EnvVar {
+	return localFileEnvVars(mmv1beta.DefaultLocalFilePath)
+}
+
+func (fs *ExternalVolumeFileStore) InitContainers(_ *mmv1beta.Mattermost) []corev1.Container {
+	return []corev1.Container{}
+}
+
+func (fs *ExternalVolumeFileStore) Volumes(mm *mmv1beta.Mattermost) ([]corev1.Volume, []corev1.VolumeMount) {
```
fmartingr, line 52: Is this used somewhere else apart from tests?

fmartingr (approved):
LGTM, just one question regarding the new fileStore data structure

apis/mattermost/v1beta1/mattermost_types.go
```diff
@@ -364,6 +367,15 @@ type ExternalFileStore struct {
 	Secret string `json:"secret,omitempty"`
 }
 
+// ExternalVolumeFileStore defines the configuration of an externally managed
+// volume file store.
+type ExternalVolumeFileStore struct {
+	// The name of the externally managed volume to be mounted for file storage.
+	VolumeName string `json:"volumeName,omitempty"`
```
mirshahriar, line 374: Do we require the `VolumeName` parameter? We can rely on `VolumeClaimName`, and if the status is `Bound`, it is safe to mount the volume. We can set `external-volume` as `Volume` name in Deployment.

pkg/mattermost/file_store.go
mirshahriar: This is used [here](https://github.com/mattermost/mattermost-operator/blob/master/pkg/mattermost/mattermost_v1beta.go#L270)

gabrieljackson: Yep, exactly. The interface provides the volume config for the deployment when it is generated.

apis/mattermost/v1beta1/mattermost_types.go
gabrieljackson: Hmm interesting. You are right that we probably don't explicitly need a separate name value. Let me think about whether it's a good idea to keep it or not. @fmartingr have any input here?

pkg/mattermost/file_store.go
fmartingr: Ok, I was searching in my supposed local master branch and couldn't find any references and I found that confusing. Thanks for the clarification.

apis/mattermost/v1beta1/mattermost_types.go
mirshahriar: PV is usually managed by storageclass. As a result, the name is entirely arbitrary. Thats why, I believe it will be difficult to use the same manifest several times in several places.

gabrieljackson: > Thats why, I believe it will be difficult to use the same manifest several times in several places.

Sorry could you help me understand what you mean by this? Are you saying that by allowing for arbitrary names there is more chance to pass an incorrect value into a resource manifest?

mirshahriar: When we create a PVC, a PV with a random name is created automatically. Lets say we have a `Mattermost` manifest like below

```
fileStore:
  externalVolume:
    volumeClaimName: mm-data
    volumeName: pvc-<random string>
```

If we want to provide `volumeName`, we need to change this field every time for a new installation.  That means, If we use same manifest multiple times without setting `volumeName` to the correct one, yes, It will hold incorrect value.

gabrieljackson: Okay I think I understand now. The intent of this filestore option is to provide the most flexibility possible for admins who need to deploy Mattermost with our non-S3 options.

> If we want to provide volumeName, we need to change this field every time for a new installation

Yeah, that's correct. I am less concerned about needing to provide specific configuration for each installation as much of our configuration is like that and the Cloud Provisioner handles that for us for most stuff. That said, we won't be actively using this filestore option ourselves and is primarily made for admins using the operator in other environments. In those cases they would only need to setup and configure everything once so it shouldn't be an issue.

All that said, I am generally in favor of minimal configuration so let's remove the volume name for now. We can always add it back later if we need to. 👍

mirshahriar (approved):
LGTM

PR #320, opened 2022-12-07
feat(filestore): use service account for s3 api calls

### LINKED ISSUE #315: Mattermost Operator Does Not Support AWS IAM Profile for S3 Bucket Auth
#### Summary

Mattermost Operator expects to have Access key and Secret key set in a secret. 
#### Expected behavior

Expected behavior is to allow Mattermost to use the IAM Profile attached to the node instance

When I try not to set the Access Key and Secret Key, it gives an error saying the secret is required. 

#### Possible fixes

Make the secret an optional part of the Mattermost object

Hunter-Thompson:
#### Summary

Makes using a secret for S3 optional. Instead, gives an option to UseServiceAccount instead.
This feature is quite important to ensure security standards.

#### Ticket Link
Fixes #315 

#### Release Note

```release-note
Makes using a secret for S3 calls optional, gives an option to UseServiceAccount with IAM role instead.
```

### DIFF
--- a/controllers/mattermost/mattermost/file_store.go
+++ b/controllers/mattermost/mattermost/file_store.go
@@ -36,7 +36,7 @@ func (r *MattermostReconciler) checkFileStore(mattermost *mmv1beta.Mattermost, r
 
 func (r *MattermostReconciler) checkExternalFileStore(mattermost *mmv1beta.Mattermost, reqLogger logr.Logger) (mattermostApp.FileStoreConfig, error) {
 	if mattermost.Spec.FileStore.External.UseServiceAccount {
-		return mattermostApp.NewExternalFileStoreInfoWithoutSecret(mattermost)
+		return mattermostApp.NewExternalFileStoreInfo(mattermost, nil)
 	}
 
 	secret := &corev1.Secret{}
@@ -46,7 +46,7 @@ func (r *MattermostReconciler) checkExternalFileStore(mattermost *mmv1beta.Matte
 		return nil, err
 	}
 
-	return mattermostApp.NewExternalFileStoreInfo(mattermost, *secret)
+	return mattermostApp.NewExternalFileStoreInfo(mattermost, secret)
 }
 
 func (r *MattermostReconciler) checkExternalVolumeFileStore(mattermost *mmv1beta.Mattermost, reqLogger logr.Logger) (mattermostApp.FileStoreConfig, error) {
--- a/controllers/mattermost/mattermost/file_store.go
+++ b/controllers/mattermost/mattermost/file_store.go
@@ -35,7 +35,7 @@ func (r *MattermostReconciler) checkFileStore(mattermost *mmv1beta.Mattermost, r
 }
 
 func (r *MattermostReconciler) checkExternalFileStore(mattermost *mmv1beta.Mattermost, reqLogger logr.Logger) (mattermostApp.FileStoreConfig, error) {
-	if mattermost.Spec.FileStore.IsExternal() && mattermost.Spec.FileStore.External.UseServiceAccount {
+	if mattermost.Spec.FileStore.External.UseServiceAccount {
 		return mattermostApp.NewExternalFileStoreInfoWithoutSecret(mattermost)
 	}
 
--- a/controllers/mattermost/mattermost/file_store.go
+++ b/controllers/mattermost/mattermost/file_store.go
@@ -2,6 +2,7 @@ package mattermost
 
 import (
 	"context"
+	"fmt"
 
 	mattermostMinio "github.com/mattermost/mattermost-operator/pkg/components/minio"
 	minioOperator "github.com/minio/minio-operator/pkg/apis/miniocontroller/v1beta1"
@@ -11,6 +12,7 @@ import (
 	mattermostApp "github.com/mattermost/mattermost-operator/pkg/mattermost"
 	"github.com/pkg/errors"
 	corev1 "k8s.io/api/core/v1"
+	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/api/resource"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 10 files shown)


mirshahriar:
How is ServiceAccount used? Not seeing that ServiceAccount has been mounted in deployment. Can you point out?

Hunter-Thompson:
> How is ServiceAccount used? Not seeing that ServiceAccount has been mounted in deployment. Can you point out?

https://github.com/mattermost/mattermost-operator/blob/ce3d25da3cdccee03f2b40f4fa90480998578ab2/pkg/mattermost/mattermost_v1beta.go#L376

Hunter-Thompson:
> > > How is ServiceAccount used? Not seeing that ServiceAccount has been mounted in deployment. Can you point out?
> > 
> > 
> > https://github.com/mattermost/mattermost-operator/blob/ce3d25da3cdccee03f2b40f4fa90480998578ab2/pkg/mattermost/mattermost_v1beta.go#L376
> 
> This is for the Kubernetes Native ServiceAccount. I'm not sure if it serves our needs. As far I know, this ServiceAccount is used to access Kubernetes API server (provides RBAC permissions) and related stuffs.

@mirshahriar, AWS provides a way to attach an IAM role to a Kubernetes native ServiceAccount. Allowing a deployment to access AWS services like S3, without a secret.

https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html

Hunter-Thompson:
The user will have to:

1. Create SA manually
2. Add annotation with IAM role
3. Deploy MM operator

MM operator wont create the SA, since it already exists.

mirshahriar:
> @mirshahriar, AWS provides a way to attach an IAM role to a Kubernetes native ServiceAccount. Allowing a deployment to access AWS services like S3, without a secret.
> 
> https://docs.aws.amazon.com/eks/latest/userguide/iam-roles-for-service-accounts.html

Thank you very much. I was unaware of that.

mirshahriar:
@Hunter-Thompson, One more thing. Can we add a check to verify whether the ServiceAccount has a certain annotation required for AWS IAM?

controllers/mattermost/mattermost/file_store.go
```diff
@@ -35,14 +35,18 @@ func (r *MattermostReconciler) checkFileStore(mattermost *mmv1beta.Mattermost, r
 }
 
 func (r *MattermostReconciler) checkExternalFileStore(mattermost *mmv1beta.Mattermost, reqLogger logr.Logger) (mattermostApp.FileStoreConfig, error) {
+	if mattermost.Spec.FileStore.External.UseServiceAccount {
+		return mattermostApp.NewExternalFileStoreInfo(mattermost, nil)
```
mirshahriar, line 39: I think you want to check **ServiceAccount** here before calling `NewExternalFileStoreInfo`, like what is done for **Secret**.

Like, 
- Get the ServiceAcount
- Check Annotation

mirshahriar:
>
>  Im skipping creating the ServiceAccount if `UseServiceAccount` is true.
> 

I agree. Added a comment [here](https://github.com/mattermost/mattermost-operator/pull/320/files#r1043517266).

mirshahriar:
> We can give the user an option to add an Annotation through the spec

For the time being, we can leave it up to the user to manually put this in annotation.

Hunter-Thompson:
@mirshahriar changed the layout like you requested.

controllers/mattermost/mattermost/mattermost.go
```diff
@@ -142,7 +143,12 @@ func (r *MattermostReconciler) checkMattermostRBAC(mattermost *mmv1beta.Mattermo
 }
 
 func (r *MattermostReconciler) checkMattermostSA(mattermost *mmv1beta.Mattermost, reqLogger logr.Logger) error {
+	if mattermost.Spec.FileStore.External != nil && mattermost.Spec.FileStore.External.UseServiceAccount {
```
fmartingr, line 146: This is my only nitpick about this PR, if we are using our own ServiceAccount and we are not creating it (which it does at the Mattermost object level), does it makes sense to specify it at the filestore level? Would it make more sense to have it on the Mattermost spec? @mirshahriar feedback on this?

fmartingr (commented):
Hey @Hunter-Thompson, this LGTM overall, I only have one nit comment below regarding the use service account flow control. Apart from that 👍

controllers/mattermost/mattermost/mattermost.go
Hunter-Thompson: I think it makes sense to have it at the external filestore level since it only affects the external file store object.

controllers/mattermost/mattermost/file_store.go
```diff
@@ -35,14 +37,30 @@ func (r *MattermostReconciler) checkFileStore(mattermost *mmv1beta.Mattermost, r
 }
 
 func (r *MattermostReconciler) checkExternalFileStore(mattermost *mmv1beta.Mattermost, reqLogger logr.Logger) (mattermostApp.FileStoreConfig, error) {
+	if mattermost.Spec.FileStore.External.UseServiceAccount {
```
Hunter-Thompson, line 40: Example over here.

controllers/mattermost/mattermost/mattermost.go
fmartingr: I'm 0/5 on this, but I usually propagate logic "up to bottom" rather than "bottom-up". That is, the service account is for Mattermost itself, not only for the filestore, so, in my mind, the definition of if it should be external or not should be done at the Mattermost level and then propagate it to the underlying requirements, rather than setting it up on a "requirement" and then act accordingly. Unsure if I'm explaining myself properly 😓 

Again, 0/5, this is not a blocker, is just something that I saw during the review, and pretty sure similar things are being done the same in other parts, that's why I wanted extra eyes on this.

fmartingr (approved):

controllers/mattermost/mattermost/mattermost.go
gabrieljackson: Yeah, I am also thinking that the service account is a more general configuration to the Mattermost app. As such it should probably be defined in the Mattermost spec section instead of the filestore even if all we use it for is the filestore for now. @fmartingr does this sound like it is lining up with your thoughts as well? I think it is, but just want to check.

fmartingr: > Yeah, I am also thinking that the service account is a more general configuration to the Mattermost app. As such it should probably be defined in the Mattermost spec section instead of the filestore even if all we use it for is the filestore for now. @fmartingr does this sound like it is lining up with your thoughts as well? I think it is, but just want to check.

Better explained in less words, yes 🙃

gabrieljackson: @Hunter-Thompson we just chatted offline as a team and have aligned on how to proceed here. We actually came to the consensus that your current implementation is fine as is. Although the service account itself if a lower level resource for the Mattermost app, the bool value for using it for filestore access makes sense to be in the filestore struct like you have it.

Thanks for your patience on this!

PR #321, opened 2022-12-08
Support PVC pending states for external volume filestores

gabrieljackson:
Pending state is valid as well as bound for PVCs that Mattermost can use. This change also corrects some SetDefaults() behavior for filestores.

Fixes https://mattermost.atlassian.net/browse/MM-48932

```release-note
Support PVC pending states for external volume filestores
```


fmartingr (approved):
LGTM

PR #323, opened 2022-12-14
Improve filestore defaults logic

gabrieljackson:
This corrects an issue where some operator managed filestore default settings could be applied when not using an operator managed filestore type. These settings would be ignored when reconciling the Mattermost resources, but could be confusing when reviewing the resource.

```release-note
Improve filestore defaults logic
```


mirshahriar (approved):
LGTM

fmartingr (approved):
👍

PR #324, opened 2022-12-15
Migrate from CircleCI to Github Actions

spirosoik:
#### Summary
We gradually moving away from CirlceCI to Github Actions as we are consolidating the CIs we use. In parallel we are introducing `trivy` in the PR level and we block the PR about the number of critical vulnerabilities.

#### Ticket Link
Ticket: https://mattermost.atlassian.net/browse/CLD-4711

#### Release Note
```release-note
NONE
```

### DIFF
--- a/.github/workflows/cd.yml
+++ b/.github/workflows/cd.yml
@@ -1,0 +1,34 @@
+name: cd
+on:
+  workflow_run:
+    workflows: ["ci"]
+    types:
+      - completed
+  push:
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
+
+permissions:
+  contents: read
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@dc323e67f16fb5f7663d20ff7941f27f5809e9b6 # v3.2.0
+        with:
+          fetch-depth: 0
+
+      - name: cd/build-docker
+        run: make build-image
+
+      - name: ci/push-docker
+        run: |
+          set -e
+          set -u
+          echo $DOCKERHUB_TOKEN | docker login --username $DOCKERHUB_USERNAME --password-stdin
+          docker tag mattermost/mattermost-operator:test mattermost/mattermost-operator:$TAG
+          docker push mattermost/mattermost-operator:$TAG
+        env:
+          TAG: ${GITHUB_REF_NAME}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,110 @@
+name: ci
+on:
+  push:
+    branches:
+      - master
+    tags:
+      - "v[0-9]+.[0-9]+.[0-9]+"
+  pull_request:
+
+permissions:
+  contents: read
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # v3.2.0
+        with:
+          fetch-depth: 0
+
+      - name: Setup Go
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
+        with:
+          go-version: "1.16.3"
+          cache: true
+
+      - name: ci/check-style
+        run: make check-style
+
+      - name: ci/check-modules
+        run: make check-modules
+
+  test:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # v3.2.0
+        with:
+          fetch-depth: 0
+
+      - name: Setup Go
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
+        with:
+          go-version: "1.16.3"
+
+      - name: Generate Operator Manifests
+        run: |
+          make clean operator-sdk
+          mkdir -p /tmp/apis/mattermost/v1alpha1
+
+          cp -R apis/mattermost/v1alpha1/* /tmp/apis/mattermost/v1alpha1
+          mkdir -p /tmp/apis/mattermost/v1beta1
+          cp -R apis/mattermost/v1beta1/* /tmp/apis/mattermost/v1beta1
+          mkdir -p /tmp/config/crd/bases
+          cp -R config/crd/bases/* /tmp/config/crd/bases
+
+          make generate manifests
+          diff /tmp/apis/mattermost/v1alpha1 apis/mattermost/v1alpha1
+          diff /tmp/apis/mattermost/v1beta1 apis/mattermost/v1beta1
+          diff /tmp/config/crd/bases config/crd/bases
+
+      - name: ci/test
+        run: make unittest goverall
+
+      - name: ci/test-e2e
+        run: ./test/e2e.sh
+        env:
+          K8S_VERSION: v1.17.5
+          IMAGE_NAME: mattermost/mattermost-operator
+          IMAGE_TAG: test
+          KIND_VERSION: v0.8.1
+          SDK_VERSION: v1.0.1
+
+  build:
+    if: ${{ github.event_name == 'pull_request' || github.ref_name  == 'master' }}
+    runs-on: ubuntu-latest
+    needs: [lint, test]
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@dc323e67f16fb5f7663d20ff7941f27f5809e9b6 # v3.2.0
+        with:
+          fetch-depth: 0
+
+      - name: Set short SHA
+        run: echo "SHORT_SHA=${GITHUB_SHA:0:7}" >> $GITHUB_ENV
+
+      - name: ci/build-docker
+        run: make build-image
+
+      - name: ci/scan-docker-security
+        uses: aquasecurity/trivy-action@9ab158e8597f3b310480b9a69402b419bc03dbd5 # v0.8.0
+        with:
+          image-ref: "mattermost/mattermost-operator"
+          format: "table"
+          exit-code: "1"
+          ignore-unfixed: true
+          vuln-type: "os,library"
+          severity: "CRITICAL,HIGH"
+
+      - name: ci/push-docker
+        run: |
+          set -e
+          set -u
+          TAG="${GITHUB_SHA:0:7}"
+          echo $DOCKERHUB_TOKEN | docker login --username $DOCKERHUB_USERNAME --password-stdin
+          docker tag mattermost/mattermost-operator:test mattermost/mattermost-operator:$TAG
+          docker push mattermost/mattermost-operator:$TAG
+        env:
+          TAG: ${GITHUB_SHA:0:7}


fmartingr:
After a lot of work @mirshahriar and I managed to get the github action working again for the manifest generation. Now the e2e tests are failing because we do not have docker-in-docker in the action. I have seen some actions around that provide kind working in actions, but I'm unsure how strict are we in using external actions in our own workflows and if those need to be checked up by security. @gabrieljackson any input on this?

fmartingr:
> After a lot of work @mirshahriar and I managed to get the github action working again for the manifest generation. Now the e2e tests are failing because we do not have docker-in-docker in the action. I have seen some actions around that provide kind working in actions, but I'm unsure how strict are we in using external actions in our own workflows and if those need to be checked up by security. @gabrieljackson any input on this?

I have created a tentative upgrade under https://github.com/mattermost/mattermost-operator/pull/329

e2e is failing on mysql... which is the same that happens to me locally (both on mac and linux) because errors on the init container of the mysql operator.

gabrieljackson:
Nice work getting this further along. I am glad that my discussion with Mir was indeed on the right track. As for the use of external actions, I am not sure if we are okay with that or not. The move it GitHub actions is being spearheaded by @spirosoik so let's see what he has to say. If external actions are not allowed here then we can just copy what we need to manage it ourselves.

spirosoik:
@gabrieljackson @mirshahriar which action you want to use? Overall we want the external actions we are going to use to be officially verified by github.

fmartingr:
> @gabrieljackson @mirshahriar which action you want to use? Overall we want the external actions we are going to use to be officially verified by github.

An action to automatically install kind: https://github.com/mattermost/mattermost-operator/pull/329

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # v3.2.0
+        with:
+          fetch-depth: 0
+
+      - name: Setup Go
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
+        with:
+          go-version: "1.19"
```
phoinixgrr, line 25: Let's use the gomod file in root dir, to dynamically setup the GoLang version during CI runtime https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+      - name: ci/check-modules
+        run: make check-modules
+
+  test:
+    runs-on: ubuntu-22.04
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@755da8c3cf115ac066823e79a1e1788f8940201b # v3.2.0
+        with:
+          fetch-depth: 0
```
phoinixgrr, line 40: Not sold on this. A shallow clone should be enough for the needs of CI.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+          path: src/github.com/mattermost/mattermost-operator
+
+      - name: Setup Go
+        uses: actions/setup-go@6edd4406fa81c3da01a34fa6f6343087c207a568 # v3.5.0
+        with:
+          go-version: "1.19"
```
phoinixgrr, line 46: Let's use the gomod file in root dir, to dynamically setup the GoLang version during CI runtime https://github.com/actions/setup-go#getting-go-version-from-the-gomod-file

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+          cache-dependency-path: src/github.com/mattermost/mattermost-operator/go.sum
+
+      - name: Generate Operator Manifests
+        run: |
+          cd ${GITHUB_WORKSPACE}/src/github.com/mattermost/mattermost-operator
+
+          make clean
+          make operator-sdk
+
+          mkdir -p /tmp/apis/mattermost/v1alpha1
```
phoinixgrr, line 57: mkdirs could be concantanated in a single line with :

mkdir -p /tmp/apis/mattermost/v1alpha1 /tmp/apis/mattermost/v1beta1 /tmp/config/crd/bases

Less syscalls !

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+
+      - name: ci/test
+        run: |
+          cd ${GITHUB_WORKSPACE}/src/github.com/mattermost/mattermost-operator
+          make unittest goverall
+
+      - name: ci/test-e2e
+        run: |
+          cd ${GITHUB_WORKSPACE}/src/github.com/mattermost/mattermost-operator
+          ./test/e2e.sh
```
phoinixgrr, line 77: Suggestion: I'd prefer to have this as part of a makefile target, to decouple from hard-calling the script. 
like: `make e2e-tests`

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+
+  build:
+    if: ${{ github.event_name == 'pull_request' || github.ref_name  == 'master' }}
+    runs-on: ubuntu-latest
+    needs: [lint, test]
+    steps:
+      - name: Checkout repo
+        uses: actions/checkout@dc323e67f16fb5f7663d20ff7941f27f5809e9b6 # v3.2.0
+        with:
+          fetch-depth: 0
```
phoinixgrr, line 93: Not sold on this. A shallow clone should be enough for the needs of CI.

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+          format: "table"
+          exit-code: "1"
+          ignore-unfixed: true
+          vuln-type: "os,library"
+          severity: "CRITICAL,HIGH"
+
+      - name: ci/push-docker
+        run: |
+          set -e
+          set -u
```
phoinixgrr, line 122:
```suggestion
          set -eu
```

.github/workflows/ci.yml
```diff
@@ -0,0 +1,119 @@
...
+          echo $DOCKERHUB_TOKEN | docker login --username $DOCKERHUB_USERNAME --password-stdin
```
phoinixgrr, line 123: I'd suggest using a pinned GhAction for docker login https://github.com/docker/login-action and remove bash

Makefile
```diff
@@ -17,7 +17,7 @@ BASE_IMAGE = gcr.io/distroless/static:nonroot
 ################################################################################
 
 GO ?= $(shell command -v go 2> /dev/null)
-PACKAGES=$(shell go list ./...)
+PACKAGES=$(shell go list ./... | grep -v vendor)
```
phoinixgrr, line 20: Why is this change introduced?

scripts/install-mysql-minio.sh
```diff
@@ -4,13 +4,9 @@ DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
...
-cat <<EOF | kubectl apply -f -
-apiVersion: v1
-kind: Namespace
-metadata:
-  name: mysql-operator
-EOF
-kubectl apply -n mysql-operator -f "${DIR}"/../docs/mysql-operator/mysql-operator.yaml
+helm repo add bitpoke https://helm-charts.bitpoke.io
+helm repo update 
+helm install mysql-operator bitpoke/mysql-operator --namespace mysql-operator --create-namespace --set "extraArgs={--mysql-versions-to-image=5.7.26=percona:5.7.35}" --version v0.6.2
```
phoinixgrr, line 9: Is this correct, since we are using MysqlVersion 8.0?

phoinixgrr (commented):
Great job overall. 🚀 

One additional comment, would be to remove the return character lines (\n), between steps/jobs. But this is personal preference.

Makefile
spirosoik: it's redundant to include vendor as it's 3rd party dependency with `go mod vendor` and breaks the lint if I aremember correctly

.github/workflows/ci.yml
spirosoik: This is not part of this PR. We are just moving everything as they are, also I believe it's easier to have this repeatable for testing purposes as we have discussed in the past but it's a discussion for a later stage.

scripts/install-mysql-minio.sh
spirosoik: @toninis any ideas here?

.github/workflows/cd.yml
```diff
@@ -0,0 +1,35 @@
...
+          fetch-depth: 0
+
+      - name: cd/build-docker
+        run: make build-image
+
+      - name: ci/push-docker
+        run: |
+          set -e
+          set -u
+          echo $DOCKERHUB_TOKEN | docker login --username $DOCKERHUB_USERNAME --password-stdin
```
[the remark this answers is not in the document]
spirosoik: This is out of scope for this PR

spirosoik:
> Great job overall. 🚀
> 
> One additional comment, would be to remove the return character lines (\n), between steps/jobs. But this is personal preference.

To be honest `\n` helps me to read it easier. I was checking the webapp last time an I had a bit hard time to read each step. This is though a personal preference. @toninis any feedback here?

scripts/install-mysql-minio.sh
toninis: It's irrelevant . This fix is for the MySQL 5.7 version . We can leave this for backwards compatibility if we ever want to test against mysql 5.7

phoinixgrr:
> > Great job overall. 🚀
> > One additional comment, would be to remove the return character lines (\n), between steps/jobs. But this is personal preference.
> 
> To be honest `\n` helps me to read it easier. I was checking the webapp last time an I had a bit hard time to read each step. This is though a personal preference. @toninis any feedback here?

Yeap, it's personal preference. 
Non blocking. Feel free to discard.

toninis:
> > Great job overall. 🚀
> > One additional comment, would be to remove the return character lines (\n), between steps/jobs. But this is personal preference.
> 
> To be honest `\n` helps me to read it easier. I was checking the webapp last time an I had a bit hard time to read each step. This is though a personal preference. @toninis any feedback here?

Yeah it depends on the person . I also like the newlines , it's easier to read . We do not mind for the bigger files on workflows in any case .

phoinixgrr (approved):

spirosoik:
@gabrieljackson I think we are good to merge.

fmartingr:
@mirshahriar Can you take a look at https://github.com/mattermost/mattermost-operator/pull/324/commits/b8c1879dda752d68cdda6bab496c1b7e8d33f758 ?

apis/mattermost/v1alpha1/clusterinstallation_utils.go
```diff
@@ -141,6 +144,9 @@ func (db *Database) SetDefaults() {
 	if db.StorageSize == "" {
 		db.StorageSize = DefaultStorageSize
 	}
+	if db.Version == "" {
+		db.Version = DefaultDatabaseVersion
```
mirshahriar, line 148: If `db.Type` is `mysql`, then `db.Version` = `DefaultMySQLVersion`.  Any thought?

fmartingr: Oh, yeah, you're totally right. I was thinking on the mattermost spec, not on the cluster installation. Thanks for catching this.

toninis (approved):

fmartingr:
So I have made some tests and it seems that the mysql-operator doesn't upgrade versions if you already have one cluster set up, so for the users that have a dev environment using the mysql-operator this should be a no-op, cluster will remain in 5.7 (even if the resource is updated with the 8.0 version).

spirosoik:
@fmartingr do you need any help here?

fmartingr:
> @fmartingr do you need any help here?

Mysql-operator is giving problems again, and something have changed with github actions logs processing because it reaches a point where the disk is full (form our `waiting for deployment to be` log lines), which causes the workflow to fail (because the mysql-operator does not launch correctly), but you can only see that if you have the tab open while the workflow is running, if you refresh the `e2e` step doesn't show any data, nor in the archive download :/

spirosoik:
What if you enable debug logs?

spirosoik:
We need to push this a bit as we need to finalise the CirleCI effort. Let me know how we can help unblock this.

fmartingr:
> We need to push this a bit as we need to finalise the CirleCI effort. Let me know how we can help unblock this.

Hey, I'm currently with other e2e issue on another project, so I couldn't look this up. Basically now we have logs but mysql pods doesn't come up, but I didn't add the ssh step to check what's going on on the cluster in github infra. Running E2E test locally works just fine (on the mac).

fmartingr:
Everything is working except that the pipeline is failing due to the trivy scan result. I'm not sure how to make that step optional since results are being "pushed" to Github. @toninis @spirosoik any ideas here? We could also leave it as it is (so we don't forget to re-enable it) until we fix our dependencies in the operator. What do you think?

toninis:
> Everything is working except that the pipeline is failing due to the trivy scan result. I'm not sure how to make that step optional since results are being "pushed" to Github. @toninis @spirosoik any ideas here? We could also leave it as it is (so we don't forget to re-enable it) until we fix our dependencies in the operator. What do you think?

@fmartingr This is a Code Scanning result . There is an option on the repo to make this not to fail and just be informative . 
I will test this now

fmartingr:
> > Everything is working except that the pipeline is failing due to the trivy scan result. I'm not sure how to make that step optional since results are being "pushed" to Github. @toninis @spirosoik any ideas here? We could also leave it as it is (so we don't forget to re-enable it) until we fix our dependencies in the operator. What do you think?
> 
> @fmartingr This is a Code Scanning result . There is an option on the repo to make this not to fail and just be informative .
> 
> It should be ok now . @gabrieljackson is pending for review . You can re-enable this from the settings page once we fix the vulnerabilities

Thanks!

apis/mattermost/v1alpha1/clusterinstallation_types.go
```diff
@@ -239,6 +239,9 @@ type Database struct {
 	// Defines the secret to be used when performing a database restore.
 	// +optional
 	BackupRestoreSecretName string `json:"backupRestoreSecretName,omitempty"`
+	// Defines the cluster version for the database to use
+	// +optional
+	Version string `json:"version,omitempty"`
```
mirshahriar, line 244: @fmartingr, It looks like this change has nothing to do with this migration. Do you want to move it (with the relevant code) to a separate pr?

gabrieljackson (approved):
LGTM. We have a plan to make improvements to this along with preparing the operator 2.0 release.

PR #325, opened 2022-12-23
Fixed OpenApi issue

mirshahriar:
#### Summary

Updated dependencies to fix OpenApi issue

#### Release Note

```release-note
None
```


fmartingr:
Hey @mirshahriar, thanks for checking this out. Sadly this still generates incorrect openapi spec on my end. Since @gabrieljackson and you generate correct spec and Github actions and my computers does not, I think this is more related to configuration rather than content.

mirshahriar:
> Sadly this still generates incorrect openapi spec on my end

Did you try this?

```
$ go mod tidy
$ go mod vendor -v
$ make generate manifests
$ make all
```

gabrieljackson:
Hmm, so we normally try to keep our k8s libraries in sync with the latest validated cluster version we are running. It would be good to confirm if the github action is resolved by this or not.

PR #329, opened 2022-12-24
Install Kind tooling in Github Action's CI

fmartingr:
#### Summary

Install Kind tooling using [helm/kind-action](https://github.com/helm/kind-action) in https://github.com/mattermost/mattermost-operator/pull/324

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49163

#### Release Note

```release-note
none
```


.github/workflows/ci.yml
```diff
@@ -69,6 +69,11 @@ jobs:
           cd ${GITHUB_WORKSPACE}/src/github.com/mattermost/mattermost-operator
           make unittest goverall
 
+      - name: Install kind tooling
```
spirosoik, line 72: I think this is may redundant right? we already try to download kind here and then run it https://github.com/mattermost/mattermost-operator/blob/master/test/e2e.sh#L25.

PR #330, opened 2022-12-25
Kind version from sha256 instead of tag

fmartingr:
#### Summary

For security, set the kind version in the `kind.x-k8s.io/v1alpha4/Cluster` resource using the sha256 checksum instead of the image tag using two different files for the two architectures the kind team supports (`amd64` & `arm64`)

#### Ticket Link

Relates to https://mattermost.atlassian.net/browse/MM-49370

#### Release Note

```release-note
None
```


gabrieljackson (approved):
Ah nice work. Are these SHAs for the `v1.22.9` images? If so we can also bump them to a valid `v1.23.X` SHA. Either way, nice improvement.

gabrieljackson:
Just re-ran the failed e2e tests and they still failed. Let me know if you need me to re-review this if any fixes need to be made.

fmartingr:
> Ah nice work. Are these SHAs for the `v1.22.9` images? If so we can also bump them to a valid `v1.23.X` SHA. Either way, nice improvement.

Upgrading the version to `v1.23.19` makes the e2e tests fail (cause of the mysql operator...) so I'm leaving it at 1.22 for now until we figure out what to do with that.

mirshahriar (approved):

gabrieljackson (approved):

PR #332, opened 2023-01-05
[MM-49370] Fix test error caused by missing arm64 docker manifest

mirshahriar:
#### Summary

**Presslabs** does not provide `arm64` docker image. To fix this, we need to use images for `linux/x86_64` platofrm.

```
docker pull --platform=linux/x86_64 quay.io/presslabs/mysql-operator:0.4.0
kind load docker-image quay.io/presslabs/mysql-operator:0.4.0
```

But by default **Presslabs** MySQL Operator uses following image for mysql
`percona@sha256:713c1817615b333b17d0fbd252b0ccc53c48a665d4cfcb42178167435a957322`

Similarly, this version does not have `arm64` variant.  And we can't load exact version (of `linux/x86_64`) into kind.

Thats why, we need to tell mysql operator to use `percona:5.7.26` instead of `sha` so that we can load `linux/x86_64` of `percona:5.7.26` into kind.

Passing following flag in operator to override version

`--mysql-versions-to-image=5.7.26=percona:5.7.26`

#### Ticket Link

  Fixes [MM-49370](https://mattermost.atlassian.net/browse/MM-49370)

#### Release Note
```release-note
Fix test error caused by missing arm64 docker manifest
```

### DIFF
--- a/docs/mysql-operator/mysql-operator.yaml
+++ b/docs/mysql-operator/mysql-operator.yaml
@@ -4329,6 +4329,7 @@ spec:
             # connect to orchestrator on localhost
             - --orchestrator-uri=http://127.0.0.1:3000/api
             - --sidecar-image=quay.io/presslabs/mysql-operator-sidecar:0.4.0
+            - --mysql-versions-to-image=5.7.26=percona:5.7.26
         - name: orchestrator
           image: quay.io/presslabs/mysql-operator-orchestrator:0.4.0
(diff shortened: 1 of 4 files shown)


fmartingr (approved):
This looks good! I can’t try it myself right now but it seems that is fixes our current development issue. Thank you!

gabrieljackson (approved):
Great fix! 👍

PR #334, opened 2023-01-30
[MM-49782] Change default revision history limit for Mattermost deployment

fmartingr:
#### Summary

Scaled down the default revision history limit for Mattermost installations deployments from 5 to 1.

This should reconcile all installation's deployments, though no downtime is expected since it would only update the `revisionHistoryLimit` field while the replicaset controller removes old revisions. 

#### Ticket Link

https://mattermost.atlassian.net/browse/MM-49782

#### Release Note

```release-note
Changed default history limit for Mattermost installations to 1
```

### DIFF
--- a/pkg/mattermost/constants.go
+++ b/pkg/mattermost/constants.go
@@ -13,7 +13,7 @@ const (
 	// possible roll-back points.
 	// More details:
 	// https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#rolling-back-a-deployment
-	defaultRevHistoryLimit = 5
+	defaultRevHistoryLimit = 1
 	// defaultMaxUnavailable is the default max number of unavailable pods out
 	// of specified `Replicas` during rolling update.
 	// More details:


gabrieljackson:
This would definitely get us our desired behavior for cloud, but I wonder if we should leave this default as is and implement an override for revision history in the provisioner instead. My thinking here is that this default is probably a better value for all non-cloud use cases. What do you think?

mirshahriar:
Question: Why do we want to change it? And Is there any use case for older revisions?

fmartingr:
> This would definitely get us our desired behavior for cloud, but I wonder if we should leave this default as is and implement an override for revision history in the provisioner instead. My thinking here is that this default is probably a better value for all non-cloud use cases. What do you think?

You're totally right.

> Question: Why do we want to change it? And Is there any use case for older revisions?

You can see the details in the Jira ticket, basically, all revisions are stored in ETCD, and having 5 extra revisions for each installation is quickly consuming our cloud etcd storage space.

Issue #336, opened 2023-01-31
Operator does not configure port 8443 for calls


DaDummy:
#### Summary
Deployments created by mattermost operator do not appear to offer a way to configure the UDP-Port 8443 for the Calls-Feature.

#### Steps to reproduce
Deploy a mattermost instance and realize there is no way to enable a service that forwards UDP-Traffic to Port 8443.

#### Expected behavior
The CRD should allow configuring additional ports on the Deployment. I am absolutely fine if I have to set up the service and/or ingress config myself.

#### Observed behavior (that appears unintentional)
There appears to be no way to instruct the operator to set up a UDP container port 8443 on the mattermost deployment's pod spec.

#### Possible fixes
Add a `additionalPorts` list to the CRD that allows specifying additional containerPorts.

PR #337, opened 2023-01-31
[GH-336] Add ContainerPorts PodExtensions field

DaDummy:
#### Summary
This PR adds a `ContainerPorts` field to the `PodExtensions` in the CRD, allowing to specify additional ports for use with plugins like the new Calls-Plugin.

#### Ticket Link
#336

#### Release Note

```release-note
Add ContainerPorts PodExtensions field to allow specifying additional custom ports to expose on the mattermost Deployment's pod spec. (e.g. 8443 for the Calls Plugin)
```


DaDummy:
Something appears to be wrong with `make generate manifests` as it generates unrelated changes when invoked as documented.

Are there any special environment variables that need to be set for the paths to start with `gitlab.com` rather than `./`?

gabrieljackson:
Thanks for the contribution. @mirshahriar and @fmartingr can one of you assist with troubleshooting the the manifest issue?

mirshahriar:
> Are there any special environment variables that need to be set for the paths to start with `gitlab.com/` rather than `./`?

@DaDummy, You need to clone this repository under `$GOPATH/src/github.com/mattermost`.  Can you verify this?

DaDummy:
Ah I see.

It might be worth mentioning in the `README.md` that the repository location is relevant and that users who do not have `GOPATH` set on their systems should set it to its implicit default of `$HOME/go`. :)

Diffs look good now from my perspective. :)

mirshahriar (commented):
@DaDummy, is it not necessary to map service ports to these container ports? If we add some ports to containers, I believe we should point the LoadBalancer Service or an Internal Service to those ports. @gabrieljackson what do you think?

DaDummy:
To be more clear:

Currently I would have to patch the mattermost `Deployment` object after every reconcile to add custom container ports matching my configuration. There is no other option. This MR is about adding another option so tinkering becomes less tedious for the time where while proper support might either be missing or would not fit the local setup like a `LoadBalancer`-based solution would in our case.

This is basically just like the custom init-containers an addition to allow custom setups. This is intended for those cases where official support is lagging behind or might not be added at all.

If you are fine with calls being implemented via `nodePort` I can try to additionally contribute dedicated support for the calls feature in a separate MR, where the operator also creates the appropriate `Service` object. Support for the `LoadBalancer`-variant however would need to be implemented by someone operating a cluster that supports those as I couldn't test that.

.gitlab-ci.yml
```diff
@@ -0,0 +1,3 @@
+include:
```
fmartingr, line 1: Please remove this file from the PR.

fmartingr (commented):
Hey @DaDummy, thanks for your contribution! I've only reviewed the code and it seems about right, only warned about a gitlab-ci file that shouldn't be there. Regarding usage, I assume you already know but you can use [resource patching](https://github.com/mattermost/mattermost-operator/blob/master/apis/mattermost/v1beta1/mattermost_types.go#L143-L150 with the operator manifest to apply custom changes to resources directly.

.gitlab-ci.yml
DaDummy: My bad - the corresponding commit wasn't supposed to land in github. Removed.

DaDummy:
> It's experimental and a last-case scenario in case you cannot setup your installation the way you want.
> 
> Something like this (from memory):
> 
> ```go
> ResourcePatch: &ResourcePatch{
>   Deployment: &Patch{
>     Patch: `[{"op": "add", "path": "/spec/template/spec/containers/0/ports/-", "value": {"containerPort": 9999}}]`
>   }
> }
> ```
> 
> Refs: [kubectl patch](https://kubernetes.io/docs/tasks/manage-kubernetes-objects/update-api-object-kubectl-patch/) | [json patch](https://jsonpatch.com/)

I see. Though as long as this is experimental and as long as `podExtensions` is not deprecated, I'd still be in favor to have this PR merged for having a less experimental approach for the custom ports use case :)

Regarding the CI-Failure: The error does not appear to be related to the changes in this PR. Do you agree?

fmartingr:
> I see. Though as long as this is experimental and as long as `podExtensions` is not deprecated, I'd still be in favor to have this PR merged for having a less experimental approach for the custom ports use case :)

Yeah, I'm not against this changes. Just an FYI if you needed to do something else.

> Regarding the CI-Failure: The error does not appear to be related to the changes in this PR. Do you agree?

It seems all PRs are failing right now, we are in the middle of a migration and something may be missconfigured.

DaDummy:
I've synced the branch with master and the CI appears to pass now. Is there anything left to do from my side?

gabrieljackson (approved):
LGTM!

fmartingr (approved):

mirshahriar (approved):

fmartingr:
Hey @DaDummy, thanks for your contribution and addressing our comments in such a timely manner, it took some more time for us to properly review with other things going on but I'm sending this to upstream now. Expect it on the next operator release :)

DaDummy:
Is this change included in v1.20.0 or will I have to wait till v1.20.1 till I can switch back to the upstream operator image?

DaDummy:
It is indeed not included. Well I'm looking forward to the next release then. Thanks for the support and merge :)

tculp:
@fmartingr Hey, I noticed this feature is not included in the latest release. Can a new release be cut which includes these changes? Thanks!

fmartingr:
> @fmartingr Hey, I noticed this feature is not included in the latest release. Can a new release be cut which includes these changes? Thanks!

Apologies, last release was to create a hotfix due to some problems with the operator and we haven't make any releases since. I will chime in the team here to know when to expect a new operator release.

cc: @gabrieljackson @nickmisasi

PR #338, opened 2023-02-02
Update dev size resource values

gabrieljackson:
These values were changed to be more similar to the cloud sizes.

```release-note
Update dev size resource values
```


mirshahriar (approved):

fmartingr (approved):

stylianosrigas (approved):
Thanks for the quick change, LGTM @gabrieljackson 🚀

PR #358, opened 2023-07-05
fix(postgres): use -h instead of --dbname when executing pg_isready

ThamosIO:
#### Summary

This replaces the use of --dbname when executing `pg_isready --dbname="$DB_CONNECTION_CHECK_URL"`since it doesn’t work. It should be `pg_isready -h $DB_CONNECTION_CHECK_URL`.

#### Release Note

```release-note
NONE
```

### DIFF
--- a/pkg/mattermost/database_external.go
+++ b/pkg/mattermost/database_external.go
@@ -110,7 +110,7 @@ func getDBCheckInitContainer(secretName, dbType string) *corev1.Container {
 			Env:             envVars,
 			Command: []string{
 				"sh", "-c",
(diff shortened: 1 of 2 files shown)


gabrieljackson:
Hey @ThamosIO, sorry about the delay on this. The existing command should still work just fine.

From the docs:
> -d dbname
--dbname=dbname
Specifies the name of the database to connect to. The dbname can be a connection string. If so, connection string parameters will override any conflicting command line options.

https://www.postgresql.org/docs/13/app-pg-isready.html

I also just tested it and got the expected behavior. I will leave this PR open for a few days for comment if you have any.

Issue #359, opened 2023-07-29
ARM64 image is not working


this-is-tobi:
#### Summary

Image for `arm64` architecture is not working.

#### Steps to reproduce

1. Add Mattermost Helm repo : 

```sh
helm repo add mattermost https://helm.mattermost.com
helm repo update
```

2. Deploy the operator on `arm64` architecture with Helm :

```sh
helm upgrade \
  --install \
  --create-namespace \
  --namespace mattermost \
  --set-string mattermostOperator.image.tag="master" \
  mattermost-operator mattermost/mattermost-operator
```

> *Notes: The image tagged `master` is available for both `amd64` and `arm64` (cf. https://hub.docker.com/layers/mattermost/mattermost-operator/master/images/sha256-46f4463f4fa537f0721b289b2430df1870c816ffdda554f75b699fcccfc04f27).*

Version used :
- Mattermost operator image tagged `master`
- Mattermost not deployed at the time

Environment : 
- Hosts 
  - Raspberry Pi `4 model B - 8GB`
  - Kernel version `6.1.21-v8+`
  - OS `Debian GNU/Linux 11 (bullseye)`
- Softwares
  - K3S version `v1.27.3+k3s1`

#### Expected behavior

Because docker image is now build for `arm64` architecture the mattermost-operator container should be able to start in that kind of hosts as it does for `amd64`.

#### Observed behavior (that appears unintentional)

```sh
kubectl -n mattermost get pod                                          
NAME                                   READY   STATUS             RESTARTS       AGE
mattermost-operator-57649f4c95-pl4wv   0/1     CrashLoopBackOff   15 (67s ago)   53m
```

```sh
kubectl -n mattermost describe pod/mattermost-operator-57649f4c95-pl4wv                               

Name:             mattermost-operator-57649f4c95-pl4wv
Namespace:        mattermost
Priority:         0
Service Account:  mattermost-operator
Node:             pi3/192.168.0.103
Start Time:       Sat, 29 Jul 2023 15:43:30 +0200
Labels:           app.kubernetes.io/instance=mattermost-operator
                  app.kubernetes.io/managed-by=Helm
                  app.kubernetes.io/name=mattermost-operator
                  helm.sh/chart=mattermost-operator-0.3.5
                  name=mattermost-operator
                  pod-template-hash=57649f4c95
Annotations:      <none>
Status:           Running
IP:               10.42.0.20
IPs:
  IP:           10.42.0.20
[...]

PR #360, opened 2023-08-03
feat: allow customizable target arch/platform on make build command

this-is-tobi:
#### Summary

This PR fix arm64 build.

*__Note:__ cf. https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope.*

#### Ticket Link

Fixes https://github.com/mattermost/mattermost-operator/issues/359.

#### Release Note

```release-note
support arm64 builds/binaries
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,6 +4,9 @@ ARG BASE_IMAGE=gcr.io/distroless/static:nonroot
 
 FROM ${BUILD_IMAGE} as builder
 
+ARG TARGETARCH
+ARG TARGETOS
+
 WORKDIR /workspace
 COPY . .
 
@@ -11,7 +14,7 @@ RUN mkdir -p licenses
 COPY LICENSE /workspace/licenses
 
 # Build
-RUN make build
+RUN make build TARGET_OS=$TARGETOS TARGET_ARCH=$TARGETARCH
 
 FROM ${BASE_IMAGE}
 
--- a/Makefile
+++ b/Makefile
@@ -28,6 +28,8 @@ MACHINE = $(shell uname -m)
 GOFLAGS ?= $(GOFLAGS:) -mod=vendor
 BUILD_TIME := $(shell date -u +%Y%m%d.%H%M%S)
 BUILD_HASH := $(shell git rev-parse HEAD)
+TARGET_OS ?= linux
+TARGET_ARCH ?= amd64
 
 BUNDLE_IMG ?= controller-bundle:$(VERSION) # Default bundle image tag
 CRD_OPTIONS ?= "crd" # Image URL to use all building/pushing image targets
@@ -119,7 +121,7 @@ goverall: $(GOVERALLS_GEN) ## Runs goveralls
 
 build: ## Build the mattermost-operator
 	@echo Building Mattermost-operator
-	GO111MODULE=on GOOS=linux GOARCH=amd64 CGO_ENABLED=0 $(GO) build $(GOFLAGS) -gcflags all=-trimpath=$(GOPATH) -asmflags all=-trimpath=$(GOPATH) -a -installsuffix cgo -o build/_output/bin/mattermost-operator $(GO_LINKER_FLAGS) ./main.go
+	GO111MODULE=on GOOS=$(TARGET_OS) GOARCH=$(TARGET_ARCH) CGO_ENABLED=0 $(GO) build $(GOFLAGS) -gcflags all=-trimpath=$(GOPATH) -asmflags all=-trimpath=$(GOPATH) -a -installsuffix cgo -o build/_output/bin/mattermost-operator $(GO_LINKER_FLAGS) ./main.go
 
 .PHONE: buildx-image
 buildx-image:  ## Builds and pushes the docker image for mattermost-operator


stafot:
@this-is-tobi THis proposal LGTM, although I slightly prefer the way it is done in other mattermost repos. 
1) We used `docker/setup-buildx-action` instead of qemu
2) Built multiarach binaries as seperate make command `make binaries` f.e https://github.com/mattermost/elrond/blob/main/Makefile#L57-L96

Are you able to refactor this PR likewise?

stafot:
Minor. Let's change the release note as it is not a bug but an enhancement to support multiarch. f.e `Support arm64 builds/binaries`

this-is-tobi:
Hi @stafot, thanks for the comment!

I may try a refactor as you mentioned but it raises 2 questions for me:
- I always use both `docker/setup-buildx-action` and `docker/setup-qemu-action` to have a diverse range of target emulations (including but not limited to `arm64`) as suggested in the [docker documentation](https://docs.docker.com/build/building/multi-platform/#building-multi-platform-images).
- In the Elrond example you sent, we can see at [line 68](https://github.com/mattermost/elrond/blob/main/Makefile#L68) that the build command uses a hard-coded `GOARCH=amd64` parameter and this is the command used in the [Dockerfile](https://github.com/mattermost/elrond/blob/main/build/Dockerfile#L9) so for me this repository is also concerned with `bad binary architecture inside the docker image` because the image (even if the target platform is set to `linux/arm64`) will always build and include binaries for `amd64`.

A proposal for this PR would be : 
1. Create a dedicated `binaries` command in the makefile to be able to publish different binaries (perhaps you'd like me to work on a deeper refactor of the publishing script to achieve harmonization with Elrond? But I'd rather do that in another PR and would probably need more information on the publishing system/process.).
2. Let the `Dockerfile` define the right `OS` and `ARCH` to build the right binary of the current target image *(i.e `linux/amd64` and `linux/arm64`)* with the [current build command proposal](https://github.com/mattermost/mattermost-operator/pull/360/files#diff-76ed074a9305c04054cdebb9e9aad2d818052b07091de1f20cad0bbac34ffb52R124).

Let me know what you think.

stafot:
@this-is-tobi You are correct about 

> In the Elrond example you sent, we can see at [line 68](https://github.com/mattermost/elrond/blob/main/Makefile#L68) that the build command uses a hard-coded GOARCH=amd64 parameter and this is the command used in the [Dockerfile](https://github.com/mattermost/elrond/blob/main/build/Dockerfile#L9) so for me this repository is also concerned with bad binary architecture inside the docker image because the image (even if the target platform is set to linux/arm64) will always build and include binaries for amd64. 
A similar but correct approach is here -> https://github.com/mattermost/mattermost-cloud-database-factory/blob/master/build/Dockerfile#L9 

For now I am ok with your current PR (Option 2) If you are willing in a future PR you can proceed with Option 1.

Silvest89:
What's the status of this @stafot ?

stafot (approved):

this-is-tobi:
Hi,

I tried to investigate on a deeper refactor to follow others Mattermost packages release system but unfortunately I think I need more informations about your release process (do you trigger the release manually or by pushing in main branch ? do you want to put binaries in the github release ? etc...). Also I just begin to develop with golang and I probably need to take a deep dive into golang release options (flags, etc...).
I hope we can merge this PR as a workaround until a more thorough overhaul of the publishing process can be done properly.

Finally, merge this PR will make the `mattermost-operator` image multi-arch but there is still the same problem for the `mattermost` server image and when I'm looking to the [repo](https://github.com/mattermost/mattermost) I don't see anything matching the production image release system (cf. [this issue](https://github.com/mattermost/mattermost/issues/23387)) like this is build from another source and I can't find any informations about this.

fmartingr:
I think we do not need qemu a this stage since we don't use CGO in the operator (neither in the server IIRC) so we can rely on [cross-compilation](https://docs.docker.com/build/building/multi-platform/#cross-compilation) directly, letting the go compiler build the binaries for each arch. As @stafot suggested using buildx should be enough.

Since this PR specifies arm64, cross-compilation should be enough for that and the default amd64 binaries.

this-is-tobi:
QEMU action removed @stafot @fmartingr

fmartingr (approved):
This LGTM, it allows users to compile to their preferred platform by cloning the repository. 

We would need to talk if we want to push/support multi-arch images ourselves, I'll leave that to @mattermost/cloud to decide.

gabrieljackson:
This is great! Thanks for the contribution @this-is-tobi.

this-is-tobi:
> This LGTM, it allows users to compile to their preferred platform by cloning the repository.
> 
> We would need to talk if we want to push/support multi-arch images ourselves, I'll leave that to @mattermost/cloud to decide.

@gabrieljackson I guess it will be done by the CI/CD : 
1. The [ci.yml](https://github.com/mattermost/mattermost-operator/blob/master/.github/workflows/cd.yml#L36) pipeline will trigger the makefile with `buildx-image` command
2. The [makefile](https://github.com/mattermost/mattermost-operator/blob/master/Makefile#L129) will trigger the `build_image.sh` script
3. The [build_image.sh](https://github.com/mattermost/mattermost-operator/blob/master/scripts/build_image.sh#L14) will trigger the build with args `"--platform" "linux/amd64,linux/arm64" "--push"`
4. The [dockerfile](https://github.com/mattermost/mattermost-operator/blob/master/Dockerfile#L17) will trigger the `make` command with architecture auto-detection and pass them as args to the makefile `build` command, see: 
- https://github.com/mattermost/mattermost-operator/blob/master/Dockerfile#L7
- https://docs.docker.com/engine/reference/builder/#automatic-platform-args-in-the-global-scope 
5. Finally the [build](https://github.com/mattermost/mattermost-operator/blob/master/Makefile#L124) command will build to the appropriate architecture

We've come full circle, even if it's not the easiest

PR #363, opened 2023-09-05
Bump Go to version 1.21

gabrieljackson:
Fixes https://mattermost.atlassian.net/browse/CLD-6261

```release-note
Bump Go to version 1.21
```


nickmisasi (approved):

fmartingr (approved):
LGTM.

It would be nice if we could leverage the `Makefile` `GOLANG_VERSION` variable into other places where is needed, to avoid having to find&replace every once in a while.

PR #367, opened 2023-11-14
Ingress deletion by label instead of name

kaiivoschneider:
#### Summary

The Mattermost operator keeps deleting Ingresses it did not create itself if they have the same name. It makes more sense for the operator to find resources by their labels. That is what Kubernetes labels are for.

#### Release Note

```release-note
NONE
```

### DIFF
--- a/pkg/resources/create_resources.go
+++ b/pkg/resources/create_resources.go
@@ -16,6 +16,7 @@ import (
 	networkingv1 "k8s.io/api/networking/v1"
 	k8sErrors "k8s.io/apimachinery/pkg/api/errors"
 	v1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/labels"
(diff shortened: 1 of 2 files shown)


pkg/resources/create_resources.go
```diff
@@ -210,17 +211,21 @@ func (r *ResourceHelper) CreatePvcIfNotExists(owner v1.Object, pvc *corev1.Persi
...
+
+	err := r.client.List(context.TODO(), foundIngresses, client.InNamespace(nameSpace), &client.ListOptions{
+		LabelSelector: labels.SelectorFromSet(labelSet),
+	})
+
+	if err != nil && len(foundIngresses.Items) > 0 {
 		return nil
 	} else if err != nil {
 		return errors.Wrap(err, "failed to check if ingress exists")
 	}
```
fmartingr, line 225: Since only one result is expected, should we handle and raise an error if the list retrieves more than one result or should we remove all results? cc: @mattermost/cloud

PR #374, opened 2024-03-25
Remove vendor directory

gabrieljackson:
This change removes the `vendor` directory in favor of pulling down dependencies as needed. The code-generator script has been moved from the `vendor` directory to `scripts` as a temporary solution until a newer process is implemented for generation.

Fixes https://mattermost.atlassian.net/browse/CLD-7371

```release-note
None
```


fmartingr (approved):
LGTM. You can remove the first line of the `.gitattributes` file too.

PR #385, opened 2024-10-29
allow setting a custom liveness probe timeout

swills:
Currently you can set some custom values for liveness probes but you can't set the timeout. I ran into an issue and needed to set it.


nickmisasi (approved):
LGTM! Thanks @swills

gabrieljackson (approved):
Thanks for the contribution!

PR #401, opened 2025-04-04
[GH-389] Usage of pinned tags fixed

### LINKED ISSUE #389: Image pinning is broken
The helm chart supports image pinning and it's part of the default [values.yaml](https://github.com/mattermost/mattermost-helm/blob/master/charts/mattermost-team-edition/values.yaml#L6).

However, doing the same with the Operator breaks the update checker.

To reproduce:

1. Set `spec.version: 10.4.2` for the Mattermost resource; observe that it starts up fine
2. Set `spec.version: 10.4.2@sha256:f655fae92b8126e7d7f5dcbb796a8bbe32883f0af9cd20841c1be035e241052f `
3. Observe that the update checker fails to start up because the Operator sets an invalid image of `mattermost/mattermost-team-edition@10.4.2@sha256:f655fae92b8126e7d7f5dcbb796a8bbe32883f0af9cd20841c1be035e241052f`

The first `@` should actually be a colon.

aboukhal:
#### Summary
Deployment was broken when using a pinned tag instead of only a tag or only a digest.

For example when using a "tag" of `1.2.3@sha26:0000000000...`, the operator would incorrectly combine this with the image name to be `...mattermost@1.2.3@sha256:0000...` instead of using the colon. This pattern of using pinned tags is commonly used in conjunction with dependency update tools like renovate or dependabot.

#### Ticket Link
Fixes #389

#### Release Note
```release-note
The operator now supports the usage of tags in the CRD that consist of tag and digest. Previously only tag or digest was supported, not both.
```

### DIFF
--- a/apis/mattermost/v1alpha1/clusterinstallation_test.go
+++ b/apis/mattermost/v1alpha1/clusterinstallation_test.go
@@ -163,6 +163,13 @@ func TestClusterInstallation(t *testing.T) {
 		assert.Contains(t, ci.GetImageName(), ci.Spec.Version)
 		assert.Equal(t, ci.GetImageName(), fmt.Sprintf("%s@%s", ci.Spec.Image, ci.Spec.Version))
 	})
+
+	t.Run("using tag and digest", func(t *testing.T) {
+		ci.Spec.Version = "10.6.1@sha256:dd15a51ac7dafd213744d1ef23394e7532f71a90f477c969b94600e46da5a0cf"
+		assert.Contains(t, ci.GetImageName(), ci.Spec.Image)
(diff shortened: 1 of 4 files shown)


aboukhal:
@gabrieljackson @stafot  sorry for pinging you directly. How can I get eyes on this PR? It seems it's not being noticed.

stafot:
@aboukhal Please consider Copilot's comments.

Lichtblitz:
@stafot Looks like AI will not take over the world anytime soon :-D
The test case it requests are literally the test case above the ones it commented.

stafot (approved):

gabrieljackson (approved):
Looks good! Thanks for the contribution.

PR #456, opened 2026-04-22
External database: unmanaged mode, custom Secret key, postgresql://

### LINKED ISSUE #437: ability to specify secret and data name within secret
When initially deploying I used cloudnative-pg to deploy a postgres database.

After the postgres database is deployed there is a secret like so:
```
$ k get secrets mattermost-cluster-app -o yaml
apiVersion: v1
data:
  dbname: redacted_base64
  fqdn-jdbc-uri: redacted_base64
  fqdn-uri: redacted_base64
  host: redacted_base64
  jdbc-uri: redacted_base64
  password: redacted_base64
  pgpass: redacted_base64
  port: redacted_base64
  uri: redacted_base64
  user: redacted_base64
  username: redacted_base64
```

Via the mattermost strategy, I can specify a secret ... so I could specify the secret provided by cloudnative-pg, but I would need the ability to say to use the 'uri' value for the DB_CONNECTION_STRING value.

This would allow for a fully automated deployment, rather than to have to copy/paste the secret value from the cloudnative-pg generated secret, to the mattermost secret.

### LINKED ISSUE #438: ability to use postgresql:// in addition to postgres://
When initially deploying I used cloudnative-pg to deploy a postgres database.

cloudnative-pg provides a connection string which starts with postgresql://, it would be nice if mattermost also recognized this url schema and used it the same as postgres.  This would help with help to allow for full automation around deployment as it would eliminate the need to edit the provided connectionstring.

I'm deploying via gitops using argocd, which lets me in general just delete a namespace, and everything will come back.  Without recognizing postgresql I'll have to start looking into using an initContainer or something and perhaps add in rbac to allow for modifying secrets.  It is probably doable, but if other folks might use various ways to deploy postgres, it may be there will be lots of folks who would benefit by recognizing 'postgresql' as a valid schema

e.g.:
```
postgresql://app:passwd_redacted@mattermost-cluster-rw.mattermost:5432/app?sslmode=disable&connect_timeout=10
```

Here's what a google search says on the matter:
```
The postgres:// and postgresql:// URL schemes are functionally equivalent and both refer to the same database protocol used to connect to a PostgreSQL database server. The difference lies solely in the naming convention used in the URL scheme.

postgresql:// is the officially recognized and preferred scheme for PostgreSQL database connections, reflecting the project's official name.
 However, postgres:// is a widely accepted and commonly used alias, often favored for its brevity and ease of pronunciation.
 This usage is so prevalent that many tools, libraries, and documentation accept both schemes interchangeably.

In practice, when connecting to a PostgreSQL database, you can use either postgres:// or postgresql:// in your connection string, and the connection will be established successfully as long as the rest of the URL (host, port, database name, credentials) is correct. The choice between the two is largely a matter of preference or convention within a specific project or organization.
```

jalet:
#### Summary
Two independent improvements to external-database support, aimed at consuming
pre-existing secrets produced by other operators (e.g. cloudnative-pg).

- `spec.database.unmanaged: true` — operator skips all DB config; user supplies
  `MM_SQLSETTINGS_*` via `spec.mattermostEnv`.
- `spec.database.external.connectionStringKey` — override the Secret key the
  operator reads the connection string from (default `DB_CONNECTION_STRING`).
  When set and no `MM_SQLSETTINGS_DATASOURCE` key is present in the Secret, the
  same key is reused as the datasource source, so a single `uri` entry is
  enough.
- Accept `postgresql://` scheme in `DB_CONNECTION_CHECK_URL` validation —
  `pg_isready` accepts both URI forms; `GetTypeFromConnectionString` already
  detected both via its `postgres` prefix check.

All three are opt-in; default behavior is unchanged.

#### Ticket Link
Fixes https://github.com/mattermost/mattermost-operator/issues/437
Fixes https://github.com/mattermost/mattermost-operator/issues/438

#### Release Note
```release-note
Add two opt-in mechanisms for connecting Mattermost to pre-existing database secrets generated by other operators: spec.database.unmanaged skips all operator database configuration, and spec.database.external.connectionStringKey overrides the Secret key used to read the connection string. The postgresql:// URI scheme is now accepted alongside postgres:// in DB connection check URLs.
```


larkox:
@nickmisasi @esarafianou Friendly reminder on this.

controllers/mattermost/mattermost/database.go
```diff
@@ -18,6 +18,10 @@ import (
 func (r *MattermostReconciler) checkDatabase(mattermost *mmv1beta.Mattermost, reqLogger logr.Logger) (mattermostApp.DatabaseConfig, error) {
 	reqLogger = reqLogger.WithValues("Reconcile", "database")
 
+	if mattermost.Spec.Database.IsUnmanaged() {
+		return nil, nil
+	}
```
nickmisasi, line 23: Returning nil, nil here weakens the DatabaseConfig contract to "maybe nil" - every furture consumer will have to remember to check. I wonder if we could have checkDatabase return a no-op UnmanagedDBConfig{} with EnvVars/InitContainers set to nil - then you could remove the changes you have in GenerateDeploymentV1Beta. The nil disappears from the pipeline entirely this way

pkg/mattermost/database_external.go
```diff
@@ -184,22 +196,24 @@ func externalDBCheckInitContainer(secretName, dbType string) *corev1.Container {
...
-// appropriate/curl.
+// appropriate/curl. The env vars mirror ExternalDBConfig.EnvVars so that a
+// custom connectionStringKey (e.g. cloudnative-pg's "uri") is honored here too.
 func builtinDBCheckInitContainer(
 	mattermost *mmv1beta.Mattermost,
 	secretName string,
 	hasSeparateDatasourceKey bool,
 	timeout time.Duration,
+	connectionStringKey string,
 ) *corev1.Container {
```
nickmisasi, line 203: I know this isn't necessarily introduced by you, but this bubbles it up - 4/5 of these args are from the ExternalDBConfig - rather than adding a 5th parameter, could we refactor this to accept the *ExternalDBConfig instead?

nickmisasi (changes requested):
Thanks for opening this PR! I have left a couple things I think should be addressed, but overall LGTM.

controllers/mattermost/mattermost/database.go
jalet: Thanks, great feedback. I took a stab at what I hope you aimed for. Let me know if it aligns with what you had in mind.

jalet: @nickmisasi is this what you had in mind?

nickmisasi (approved):

PR #461, opened 2026-05-20
Update golang version to 1.26.3

esarafianou:
## Summary

Bumps Go from 1.25.9 to 1.26.3 across the operator and updates FIPS image references:

- Go toolchain bumped to `1.26.3`
- FIPS runtime image `cgr.dev/mattermost.com/glibc-openssl-fips` bumped to `16.1.0@sha256:3ffd0b8b...` 

#### Release Note
```release-note
Updates Golang to 1.26.3 and bumps FIPS base image (glibc-openssl-fips) to 16.1.0.
```

Made with [Cursor](https://cursor.com)

PR #462, opened 2026-05-20
ci: cancel superseded workflow runs and bump anchore-scan action

esarafianou:
## Summary

- **`.github/workflows/ci.yml`** — add a top-level `concurrency` block (`group: ${{ github.workflow }}-${{ github.ref }}`, `cancel-in-progress: true`). On PR #461 we observed three runs (#433/#434/#435) all in progress for the same branch; with this block, new pushes on a PR branch supersede their in-flight predecessors while pushes to `master` and tag refs remain isolated (each gets its own ref-scoped group).
- **`.github/workflows/cd.yml`** — bump `mattermost/actions/delivery/anchore-scan` from `003fac68730de6e3e2dc31939e7f2c460f2a8ba0` to `354df0f666759a18085c7c7135db533f6b1da367` (current `HEAD` of `mattermost/actions`, "Update anchorectl to match instance version (#56)", 2026-05-20). Both `anchore-operator` and `anchore-operator-fips` steps updated to keep the pinned SHA current.

```release-note
NONE
```

Made with [Cursor](https://cursor.com)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -10,7 +10,7 @@ on:
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
-  cancel-in-progress: true
+  cancel-in-progress: ${{ github.event_name == 'pull_request' }}
 
 env:
   ENABLE_FIPS_BUILDS: true # Set to false to disable FIPS builds
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,6 +8,10 @@ on:
       - "v**"
   pull_request:
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 env:
   ENABLE_FIPS_BUILDS: true # Set to false to disable FIPS builds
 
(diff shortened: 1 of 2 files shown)


.github/workflows/ci.yml
```diff
@@ -8,6 +8,10 @@ on:
       - "v**"
   pull_request:
 
+concurrency:
```
NARSimoes, line 11: Just a sanity check: with the current configuration each new push to master cancels the previous in-progress ci run on `refs/heads/master`, because they share one concurrency group and `cancel-in-progress: true`. So, this can happens if we fast merge multiple pull-request which might affect some docker pushes that we do in CI. If this is accurate I wonder if we should just cancel in pull-request (e.g. `cancel-in-progress: ${{ github.event_name == 'pull_request' }}`).

NARSimoes (commented):
Thanks @esarafianou , just a minor clarification

.github/workflows/ci.yml
esarafianou: Good catch, addressing now.

NARSimoes (approved):

PR #472, opened 2026-08-20
Prepare operator v2

andrleite:
#### Summary

> for reference: https://github.com/mattermost/mattermost-operator/pull/470 we changed base to v2 branch instead of master

Operator **v2.0.0**. Removes the long-deprecated Custom Resources and both "delegate to another operator" integrations, and brings the Kubernetes libraries up to date.

`127 files changed, 1844 insertions(+), 27169 deletions(-)`

**Removed**

| | Why |
|---|---|
| `ClusterInstallation` (`mattermost.com/v1alpha1`) | Superseded by `Mattermost`. Carried BlueGreen/Canary, a MySQL and a PostgreSQL path, MinIO provisioning, and the CI→MM conversion code. |
| `MattermostRestoreDB` (`mattermost.com/v1alpha1`) | Only ever worked against operator-managed MySQL, and only addressed `ClusterInstallation`. |
| `spec.fileStore.operatorManaged` | Created a `MinIOInstance` for the **archived** `minio/minio-operator`. See below. |
| `spec.database.operatorManaged` | Created a `MysqlCluster` for the presslabs MySQL operator, whose API types were vendored into this repo. |
| `apis/mattermost/v1alpha1` | Nothing served it once the two CRDs were gone. |

**Result:** the Operator now depends on **no third-party operator**. Its ClusterRole requests permissions only on `""`, `coordination.k8s.io`, `apps`, `batch`, `networking.k8s.io`, `rbac.authorization.k8s.io`, and `installation.mattermost.com` — both `miniocontroller.min.io` and `mysql.presslabs.org` are gone.

**Dependency upgrade** (first commit, non-breaking and reviewable on its own): `k8s.io/*` v0.33.1 → **v0.36.3**, `controller-runtime` v0.21.0 → **v0.24.1**, adding `structured-merge-diff/v6`. No source changes were needed. Rebased onto current master, so Go 1.26.6 and master's `golang.org/x/*` bumps (incl. `x/net` v0.58.0) are preserved — all newer than what the k8s bump pulled in.

**Why MinIO had to go, specifically:** `github.com/minio/minio-operator` was last published in **February 2020** and is archived. The `miniocontroller.min.io/v1beta1 MinIOInstance` API isn't served by any current MinIO operator (which moved to `minio.min.io/v2 Tenant`), so the path has been incompatible with a supported MinIO operator for years. Worse, the `MinIOInstance` was created **without an image**, so the MinIO server holding all user uploads ran whatever `minio/k8s-operator:1.0.7` defaulted to — invisible to image scanning, because no tag appeared anywhere in this repo. Removal also drops two init containers from every pod in that mode, one of which ran `appropriate/curl:latest` (abandoned image, mutable tag).

#### Commits

The removal is structured so each step can be reviewed and reverted independently:

- `f1ec29d2` Update k8s libraries to v0.36.3 and controller-runtime to v0.24.1
- `b1e32c37` Remove the MattermostRestoreDB custom resource
- `007cd488` Remove the ClusterInstallation custom resource and controller
- `7455b02c` Remove operator-managed MinIO file store
- `b6ff2036` Remove operator-managed MySQL database
- `e1a7c988` Remove the v1alpha1 API package and move size presets to v1beta1
- `d2e20ecd` Release Operator v2.0.0

Follow-ups from review and CI:

- `6a40ac16` fix: apply CodeRabbit auto-fixes
- `d794a55c` Fix shared size presets and give the e2e suite its prerequisites
- `2a65c3f2` Use the external MinIO fixture for the e2e file store
- `e83683fd` Reconcile the CodeRabbit auto-fix commit
- `e638f379` Assert replica isolation in the explicit-size test
- `836c7c3a` Update CI for the removed v1alpha1 package
- `b7754489` Wait for Postgres and MinIO before creating Mattermost resources in e2e
- `0891657c` Keep the generated CRD under the annotation size limit

#### Reviewer notes

**A database and file store are now mandatory — deliberately with no replacement default.** Until now a `Mattermost` with neither section silently received in-cluster MinIO and MySQL; such a resource is now rejected with an explicit error. Defaulting to `local` instead would have repointed an installation at an empty PVC while its files stayed in the old MinIO — data loss from the user's perspective. Failing loudly forces a conscious choice and a migration. Two example manifests relied on that silent default and have been fixed.

**Nothing deletes user data.** No code path removes a `MinIOInstance` or `MysqlCluster`. They are left running but unmanaged, so operators can migrate at their own pace. Confirmed on a real cluster — see Testing.

**Size presets shrank but behave identically.** `ClusterInstallationSize{App, Minio, Database}` → `v1beta1.Size{App}`, since the other two only sized the removed workloads. All nine size keys keep their existing app-server replicas and resource requests, so `spec.size` is unchanged.

**Correctness fix worth a look (`d794a55c`).** The size presets were aliased rather than copied: `setDefaultReplicasAndResources` took `&DefaultSize.App.Replicas`, handing every `Mattermost` a pointer into a package-level value, and `ResourceRequirements` assignment shares its `ResourceList` maps by reference. Scaling one installation therefore rewrote the preset and changed the defaults every installation reconciled afterwards received. Demonstrated before fixing — one CR moved `DefaultSize.App.Replicas` from 2 to 999 and CPU from 500m to 777m, and a second unrelated CR then inherited those. Two regression tests fail against the previous code. This predates v2 (v1alpha1 aliased identically), but the code moved to v1beta1 here.

**The 1480-line CRD diff in `0891657c` is intentional.** `make deploy` began failing with `metadata.annotations: Too long: must have at most 262144 bytes`. `kubectl apply` stores the manifest in `last-applied-configuration`, and measured as JSON — which is what the annotation holds — master sat at 260,965 bytes, **1,179 bytes under** the ceiling. The k8s v0.36 bump pushed it over by inlining new `corev1` fields; the removals recovered 6.3KB but not enough. `CRD_OPTIONS` now passes `crd:maxDescLen=200`, bringing it to 192,749 bytes, 69KB clear. **Consequence:** `kubectl explain` shows truncated descriptions for long fields; the full text remains in `docs/mattermost_v1beta1_crd.md`, which is generated from Go source and unaffected. Note master is still 1,179 bytes from this same failure and should get the same treatment independently.

**⚠️ Go API break for `mattermost-cloud`**, which imports these packages directly: `apis/mattermost/v1alpha1` is gone; size presets moved to `v1beta1.Size`; `CalculateResourceMilliRequirements`/`CalculateCPUMilliRequirement`/`CalculateMemoryMilliRequirement` lost their `includeDatabase`/`includeMinio` parameters; and `OperatorManagedMinio`, `OperatorManagedDatabase`, `pkg/components/{minio,mysql}` and `pkg/mattermost.NewOperatorManagedFileStoreInfo` are removed. **This needs a coordinated change there and is the remaining blocker.**

**Not in this PR, by design:** the health-checker label-selector bug (`MattermostPodLabels` vs `MattermostLabels`, dates to 2022, breaks any install using `podTemplate.extraLabels`); `useServiceLoadBalancer` orphaning a leftover Ingress; and bumping the e2e `K8S_VERSION` (see Testing). All pre-date or sit outside this work and deserve their own PRs.

#### Testing

**All CI checks pass.** Note `ci/test-e2e` is a step inside the `test` job rather than its own check — it stands up a kind cluster, runs `make deploy`, then runs both `./test/e2e` and `./test/e2e-external`.

**Validated manually in a real dev cluster:**
- pre-existing installations continue to reconcile and serve traffic across the upgrade
- a newly created installation comes up clean

This matters because the e2e job is pinned to `K8S_VERSION: v1.22.9`, 14 minor versions behind the v0.36.3 client libraries this PR moves to — well outside client-go's supported skew. A green e2e is therefore weaker evidence than it appears, and the dev-cluster run is what actually covers modern Kubernetes. Bumping that pin is worth a follow-up, kept out of this PR to avoid surfacing unrelated failures in an already-large change.

Locally: `go build ./...`, `gofmt`, `go vet ./...` (including both e2e packages) and `make unittest` pass **after every individual commit**. Generated artefacts verified idempotent — re-running `make generate manifests` and `kustomize build config/default` produces no diff, re-checked after the rebase onto master.

The e2e suite was reworked as part of this. It previously got its database *from the MySQL operator*; it now uses an in-cluster Postgres and a standalone MinIO (as an S3 endpoint) from `resources/`, which is what `test/e2e-external` already did. A new `SetupMattermostPrerequisites` applies both and waits for their Deployments to become available before any `Mattermost` is created — necessary because `mm-secrets.yaml` carries no `DB_CONNECTION_CHECK_URL`, so the Operator injects no database readiness init container and nothing else blocks on the database being reachable.

#### Ticket Link

SEC-11130

#### Release Note

```release-note
Mattermost Operator v2.0.0 removes the deprecated `ClusterInstallation` and `MattermostRestoreDB` custom resources, and the operator-managed database and file store options (`spec.database.operatorManaged`, `spec.fileStore.operatorManaged`). A database and a file store are now required: set `spec.database.external` plus one of `spec.fileStore.external`, `spec.fileStore.local` or `spec.fileStore.externalVolume`. action required — before upgrading: (1) if you still run `ClusterInstallation` resources, upgrade to Operator v1.24.x first and let it migrate them to `Mattermost`, since the migration code no longer exists in v2.0.0; (2) delete any remaining `MattermostRestoreDB` objects, as that CRD is removed and they become unreadable; (3) if you use `database.operatorManaged`, move the database to one you manage and point `database.external` at a secret with its connection string; (4) if you use `fileStore.operatorManaged`, copy the objects out of MinIO into an S3-compatible bucket or a PVC. Existing MysqlCluster and MinIOInstance objects are not deleted, but the Operator stops managing them and Mattermost will no longer be pointed at them. Note that `kubectl explain` now shows shortened descriptions for some fields; the full reference is in docs/mattermost_v1beta1_crd.md.
```


apis/mattermost/v1beta1/db_util.go
```diff
@@ -1,83 +1,29 @@
...
 
-// SetDefaults sets the missing values in Database to the default ones.
-func (db *Database) SetDefaults() {
+// SetDefaults validates the database configuration.
+//
+// There is deliberately no default. Until Operator v2 an unconfigured database
+// silently became a MySQL cluster provisioned through the presslabs MySQL
+// operator; that integration has been removed, so a database now has to be
+// supplied explicitly.
+func (db *Database) SetDefaults() error {
```
fmartingr: nit: This method doesn't set defaults anymore, but performs validation.

Maybe renaming it to `IsValid` ?

apis/mattermost/v1beta1/file_store_util.go
```diff
@@ -4,23 +4,27 @@
...
-	if fs.isAnyExceptOperatorManaged() {
-		return
+// SetDefaults validates the file store configuration.
+//
+// There is deliberately no default. Until Operator v2 an unconfigured file store
+// silently became an in-cluster MinIO provisioned through the MinIO operator, which
+// no longer exists. Picking a different default instead would repoint an existing
+// installation at empty storage while its files stayed in the old MinIO, so this
+// fails and asks for an explicit choice.
+func (fs *FileStore) SetDefaults() error {
```
fmartingr: nit: This method doesn't set defaults anymore, but performs validation.

Maybe renaming it to `IsValid` ?

apis/mattermost/v1beta1/mattermost_utils.go
```diff
@@ -24,7 +24,7 @@ const (
 	DefaultMattermostSize = "5000users"
 	// DefaultMattermostDatabaseType is the default Mattermost database
 	DefaultMattermostDatabaseType = "mysql"
```
fmartingr: Do we need all these defaults if we are not provisioning defaults database/storage?

apis/mattermost/v1beta1/mattermost_sizes.go
```diff
@@ -33,25 +32,26 @@ func (mm *Mattermost) SetReplicasAndResourcesFromSize() error {
...
+// of a preset field, or assigning a ResourceRequirements (whose ResourceLists are
+// maps), would let one resource's replica count or resource requests mutate the
+// preset itself and leak into every Mattermost reconciled afterwards.
+
 func (mm *Mattermost) setDefaultReplicasAndResources() {
 	mm.Spec.Size = ""
 
 	if mm.Spec.Replicas == nil {
-		mm.Spec.Replicas = &mattermostv1alpha1.DefaultSize.App.Replicas
+		mm.Spec.Replicas = utils.NewInt32(DefaultSize.App.Replicas)
```
fmartingr: I believe we can do this in go 1.26:

```suggestion
		mm.Spec.Replicas = new(DefaultSize.App.Replicas)
```

test/e2e/setup.go
```diff
@@ -62,3 +59,69 @@ func SetupTest() (TestEnvironment, error) {
 func boolPtr(b bool) *bool {
 	return &b
 }
+
+// SetupMattermostPrerequisites creates the database and file store a Mattermost
+// now requires. The Operator used to provision both itself via the MySQL and
+// MinIO operators; since it no longer does, the suite has to supply them.
```
fmartingr: nit: It feels weird when comments references things from the past.

```suggestion
// SetupMattermostPrerequisites creates the database and file store Mattermost
// requires. 
```

test/e2e/mattermost_test.go
```diff
@@ -37,19 +35,17 @@ func TestMattermost(t *testing.T) {
...
-		err = waitForDeployment(t, k8sTypedClient, "minio-operator", "minio-operator", 1, retryInterval, timeout)
-		require.NoError(t, err)
-	})
 	t.Run("mattermost operator ready", func(t *testing.T) {
 		err = waitForDeployment(t, k8sTypedClient, mmNamespace, "mattermost-operator", 1, retryInterval, timeout)
 		require.NoError(t, err)
 	})
 
+	// A database and a file store are now required and the Operator does not
+	// provision either, so the suite has to create them before any Mattermost.
```
fmartingr:
```suggestion
```

apis/mattermost/v1beta1/mattermost_sizes.go
```diff
@@ -21,7 +20,7 @@ func (mm *Mattermost) SetReplicasAndResourcesFromSize() error {
 		return nil
 	}
 
-	size, err := mattermostv1alpha1.GetClusterSize(mm.Spec.Size)
+	size, err := GetClusterSize(mm.Spec.Size)
```
fmartingr: nit: `GetClusterSize` -> `GetMattermostSize` ?

apis/mattermost/v1beta1/mattermost_utils_test.go
```diff
@@ -12,6 +12,10 @@ import (
 func TestMattermost_SetDefaults(t *testing.T) {
 	mm := &Mattermost{Spec: MattermostSpec{
 		IngressName: "",
+		// A file store is mandatory, so every fixture that expects SetDefaults to
+		// succeed has to configure one.
```
fmartingr:
```suggestion
```

apis/mattermost/v1beta1/file_store_util.go
fmartingr: We would also need to check where this method is used, because it's not the same to assume that this method will set the defaults for the Mattermost instance than validating the Mattermost instance values are correct, maybe an error here breaks something along the line.

fmartingr (commented):
Outstanding work @andrleite ! 🎊 

I would give it a pass with an agent to review all doc strings around, it seems there are some mentions to removed fields here and there and there are many mentions on how "things used to be" which are not useful moving on.

Some questions:
- Shouldn't we drop support for MySQL entirely?
- Why keeping the APIs in `v1beta`? Are there plans to move them to `v1` (or `v2`) at some point?

I've left some comments around in the changes, most of them are just small things that I leave to your judgement. The most critical for me is the code path `SetDefault` take if we are now using it for validation. I would rename the method name and make a pass to check if there's any problems with that.

Again, great job here!

test/e2e/mattermost_test.go
```diff
@@ -37,19 +35,17 @@ func TestMattermost(t *testing.T) {
...
+	cleanupPrereqs, err := SetupMattermostPrerequisites(context.TODO(), k8sClient, mmNamespace)
+	require.NoError(t, err)
+	defer cleanupPrereqs()
+
```
nickmisasi: I think this is a slight behaviour regression because prior to this change each test would have had its own db and minio, but now the scale and upgrade tests share

apis/mattermost/v1beta1/mattermost_utils.go
```diff
@@ -70,8 +70,12 @@ func (mm *Mattermost) SetDefaults() error {
...
 	}
 
-	mm.Spec.FileStore.SetDefaults()
-	mm.Spec.Database.SetDefaults()
+	if err := mm.Spec.FileStore.SetDefaults(); err != nil {
+		return err
+	}
+	if err := mm.Spec.Database.SetDefaults(); err != nil {
+		return err
+	}
```
nickmisasi: Because SetDefaults now just returns an error, an upgraded operator-managed install will only show an error for the filestore. The tests pass an empty spec and assert only the error for the filestore missing. Not sure we need to have a single error when both are missing (though it would be nice if we didn’t have to have 2 failures) but we should at least add test coverage for the case of filestore existing but no database

nickmisasi (commented):
Overall LGTM - flagged a couple things in addition to Felipe

apis/mattermost/v1beta1/mattermost_sizes.go
andrleite: you're right, let's do it

andrleite:
> Outstanding work @andrleite ! 🎊
> 
> I would give it a pass with an agent to review all doc strings around, it seems there are some mentions to removed fields here and there and there are many mentions on how "things used to be" which are not useful moving on.
> 
> Some questions:
> 
> * Shouldn't we drop support for MySQL entirely?
> * Why keeping the APIs in `v1beta`? Are there plans to move them to `v1` (or `v2`) at some point?
> 
> I've left some comments around in the changes, most of them are just small things that I leave to your judgement. The most critical for me is the code path `SetDefault` take if we are now using it for validation. I would rename the method name and make a pass to check if there's any problems with that.
> 
> Again, great job here!

Thanks @fmartingr .

Drop support for MySQL entirely, agreed. Mattermost officially dropped MySQL for new installs in v10, with full end-of-support in v11. We removed operator-managed MySQL in this PR and updated the README to recommend PostgreSQL. The external MySQL code path stays for existing users mid-migration, but a follow-up PR should add a validation warning when a MySQL connection string is detected so the deprecation is explicit.

Regarding `v1beta`, good call — it's overdue. Not in this PR since it would add Kubernetes API serving complexity on top of an already large change, and v1 carries a stability promise we want to be confident about before making. Plan is to do it in the next minor (v2.1) once v2.0.0 has had a release cycle to settle.

apis/mattermost/v1beta1/mattermost_sizes.go
andrleite: You're right — Applied, thanks for the catch!

fmartingr (approved):
👏 👏 👏 👏  Good job.

nickmisasi (approved):
LGTM

PR #473, opened 2026-08-31
Add httproute typed sdk

andrleite:
This pull request introduces support for configuring an `HTTPRoute` (Gateway API) resource in the Mattermost Operator, allowing users to use Gateway API-based routing instead of, or alongside, traditional Kubernetes Ingress resources. The changes add new CRD fields, validation logic, accessors, and comprehensive unit tests to ensure correct behavior and backward compatibility.

**HTTPRoute (Gateway API) support:**

* Added a new `HTTPRoute` field to `MattermostSpec`, with a corresponding `HTTPRouteSpec` struct that allows enabling/disabling HTTPRoute, specifying hosts, referencing a Gateway, setting annotations, and timeouts. This includes a nested `GatewayReference` type. [[1]](diffhunk://#diff-b4621be397c99475f94b000516d66b895208dcbb094ed8234d4bc03ae60b8496R82-R86) [[2]](diffhunk://#diff-b4621be397c99475f94b000516d66b895208dcbb094ed8234d4bc03ae60b8496R228-R274)
* Updated OpenAPI schema generation and deepcopy methods to support the new `HTTPRoute` and `GatewayReference` types, ensuring CRD compatibility and proper code generation. [[1]](diffhunk://#diff-877a7e4dccd6dcf8e19dd7166872a3fc03081c2f78d4df5290944058eb61a68fR206-R211) [[2]](diffhunk://#diff-877a7e4dccd6dcf8e19dd7166872a3fc03081c2f78d4df5290944058eb61a68fL342-R348) [[3]](diffhunk://#diff-2db6a61596864df29c52f0a07dc5db4113d258a4598020160c63eb88099a9ff6R214-R256) [[4]](diffhunk://#diff-2db6a61596864df29c52f0a07dc5db4113d258a4598020160c63eb88099a9ff6R444-R448)

**Validation and logic changes:**

* Enhanced `SetDefaults` to validate that when HTTPRoute is enabled, a host and gateway reference name are required, and to ensure Ingress and HTTPRoute interaction is handled correctly.
* Updated `IngressEnabled` logic to allow HTTPRoute to opt out of the default Ingress creation, preserving legacy behavior and supporting side-by-side migration scenarios.

**Accessors and helpers:**

* Added new methods to the `Mattermost` type for accessing HTTPRoute configuration, including enablement, host(s), annotations, and a unified method for determining the site URL host, which respects the new HTTPRoute logic while preserving legacy behavior.

**Unit tests:**

* Added comprehensive tests to cover HTTPRoute validation, Ingress/HTTPRoute interaction, accessors, and to ensure that legacy behavior is preserved when HTTPRoute is not used.

These changes provide a robust foundation for Gateway API support in the Mattermost Operator, while maintaining backward compatibility and a smooth migration path for existing users.

#### Release Note
```release-note
Added opt-in [Gateway API](https://gateway-api.sigs.k8s.io/) HTTPRoute support. Operators running on clusters with the Gateway API CRDs installed can now route traffic to Mattermost through a Gateway instead of an Ingress by setting httpRoute.enabled: true and referencing a Gateway:


spec:
  httpRoute:
    enabled: true
    host: mattermost.example.com
    gatewayRef:
      name: shared-gateway
      namespace: gateway-system   # optional, defaults to MM namespace
      sectionName: https          # optional
    requestTimeout: "3600s"       # optional, default
    backendRequestTimeout: "3600s"
Ingress and HTTPRoute can coexist during migration — set ingress.enabled: false explicitly to stop the Operator from creating an Ingress alongside the HTTPRoute. Clusters without the Gateway API CRDs installed are unaffected: the Operator skips HTTPRoute deletion silently on installations that never opted in.
```


rahb3rt (approved):

apis/mattermost/v1beta1/mattermost_utils.go
```diff
@@ -199,6 +219,68 @@ func (mm *Mattermost) GetIngressClass() *string {
...
+	return hosts
+}
+
+// GetHTTPRouteAnnotations returns HTTPRoute annotations.
+func (mm *Mattermost) GetHTTPRouteAnnotations() map[string]string {
+	if mm.Spec.HTTPRoute == nil {
+		return nil
+	}
+	return mm.Spec.HTTPRoute.Annotations
+}
```
nickmisasi, line 263: I think it would make more sense to move these under HTTPRoute. For example:

```
// GetHTTPRouteAnnotations returns HTTPRoute annotations.
func (mm *Mattermost) GetHTTPRouteAnnotations() map[string]string {
	if mm.Spec.HTTPRoute == nil {
		return nil
	}
	return mm.Spec.HTTPRoute.Annotations
}
```

Would become:
```
func (h *HTTPRoute) Annotations() map[string]string {
	if h == nil {
		return nil
	}
	return h.Annotations
}
```

controllers/mattermost/mattermost/mattermost.go
```diff
@@ -216,6 +225,38 @@ func (r *MattermostReconciler) checkMattermostIngress(mattermost *mmv1beta.Matte
 	return r.Resources.Update(current, desired, reqLogger)
 }
 
+func (r *MattermostReconciler) checkMattermostHTTPRoute(mattermost *mmv1beta.Mattermost, reqLogger logr.Logger) error {
+	// See checkMattermostIngress: UseServiceLoadBalancer takes precedence over any
+	// L7 routing configuration, and suppressing it means removing the resource, not
+	// merely declining to create one.
+	if mattermost.Spec.UseServiceLoadBalancer && mattermost.HTTPRouteEnabled() {
```
nickmisasi, line 232: The above suggestion would make access better too IMO. `mattermost.HTTPRoute.Enabled()`

nickmisasi (changes requested):
Overall LGTM. One refactor suggestion

fmartingr (approved):

apis/mattermost/v1beta1/mattermost_utils.go
andrleite: Hey Nick, good suggestion. One thing: Go doesn't allow a method and a field to share the same name, so Annotations() on a struct that already has an Annotations field won't compile — same for Enabled(). Two options:

Use Get prefix — GetEnabled(), GetAnnotations(), etc.
Keep the accessors on *Mattermost as-is, matching the existing Ingress pattern (GetIngressAnnotations(), GetIngressHost())
Which do you think?

nickmisasi (approved):
Thanks Andre!

